PR #94, opened 2020-12-21
Installation documentation

ToddG:
### Purpose
Add additional documentation on how to install and use `REISE.jl` with some hints on how to verify and troubleshoot that the installation has been successful.

### Where to Look
Everything is in the `README.md`.

### Time estimate
20 min

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,43 +1,236 @@
- # REISE.jl
+# REISE.jl
 Renewable Energy Integration Simulation Engine.
 
 This repository contains, in the **src** folder, the Julia scripts to run the power-flow study in the U.S. electric grid. The simulation engine relies on [Gurobi] as the optimization solver.
 
+## Table of Contents
+1. [Dependencies](#dependencies)
+2. [Installation (Local)](#installation-local)
+3. [Usage (Julia)](#usage-julia)
+4. [Usage (Python)](#usage-python)
+5. [Docker](#docker)
+6. [Package Structure](#package-structure)
+7. [Formulation](#formulation)
+
+
 ## Dependencies
-This package requires installations of the following
-- [Julia]
-- [Gurobi]
-- [Python]
+This package requires installations of the following, with recommended versions listed.
+- [Julia], version 1.5
+- [Gurobi], version 9.1
+- [Python], version 3.8
+
+This package can also be run using Docker, in which the following are required:
+- [Docker]
+- [Gurobi Cloud License] file
 
 For sample data to use with the simulation, please visit [Zenodo].
 
+### System Requirements
 
-## Installation
-### Julia package
-The most reliable way to install this package is by cloning the repo locally, navigating to the project folder, activating the project, and instantiating it. This approach will copy install all dependencies in the **exact** version as they were installed during package development. **Note**: if `Gurobi.jl` is not already installed in your Julia environment, then its build step will fail if it cannot find the Gurobi installation folder. To avoid this, you can specify an environment variable for `GUROBI_HOME`, pointing to the Gurobi `<installdir>`.
+Large simulations can require significant amounts of RAM. The amount of RAM necessary is proportional to both the size of the grid and the size of the interval with which to run the simulation.
 
-For more information, see https://github.com/JuliaOpt/Gurobi.jl#installation.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -455,9 +648,14 @@ Penalty for transmission line limit violations (if transmission violations are e
...
 [Julia]: https://julialang.org/
-[Download Julia]: https://julialang.org/downloads/
-[Zenodo]: https://zenodo.org/record/3530898
-[Python]: https://www.python.org/
\ No newline at end of file
+[Download Julia]: https://julialang.org/downloads/#current_stable_release
+[Python]: https://www.python.org/
+[Download Python]: https://www.python.org/downloads/release/python-386/
+[Docker]: https://docs.docker.com/get-docker/
+[Zenodo]: https://zenodo.org/record/4450759
```
danielolsen, line 658: Please do not update the Zenodo link. The `3530898` link will always automatically redirect to the latest version of the dataset, any other link points to a specific version.

README.md
```diff
@@ -242,10 +436,9 @@ this repository.
 
 ## Docker
 
-The easiest way to setup this engine is within a Docker image. There is an
-included `Dockerfile` that can be used to build the Docker image. With the
-Docker daemon installed and running, navigate to the `REISE.jl` folder
-containing the `Dockerfile` and build the image:
+The easiest way to setup this engine is within a Docker image. Note, however, that the Docker image uses a [Gurobi Cloud License] and may not be able to handle larger simulations.
```
kasparm, line 439: I would be bit more specific in this sentence regarding machine requirements in order to run larger simulations.
@danielolsen do you have a good formulation for that?

rouille: I agree. We probably want to say here that it requires significant resources to run a scenario in the U.S electric grid over a full year:
* x hours on y CPU
* z GB of RAM
* f GB of disk space to store the outputs of the simulation

danielolsen: I don't have a good formula. RAM requirements will vary based on the interval length, total CPU-hours should be fairly similar, outputfile sizes can be pretty precise. I'm not running any USA scenarios right now so I don't have any `top` outputs to share.

I would think that RAM requirements are the most significant constraint, and that's more about interval length than the fact that it's a Docker image or that it uses a cloud license.

danielolsen: Based on the Western scenarios that I have running now, and my recollection of the USA ones run recently, I would estimate that 1-2 GB of RAM are needed for each time interval in a USA run. E.g. a 24-hour interval simulation would require 24-48 GB; if only 16 GB RAM are available, I would recommend a time interval of 8 hours or less.

README.md
ahurli: Good to know! It wasn't clear to me which (if any) Zenodo links wouldn't become out of date.

README.md
ahurli: This makes sense. I saw the memory constraints when attempting some larger simulations in the container, but if the container/cloud architecture isn't the issue, I removed any mention of how Docker/the cloud license may constrain the simulation size.

Instead, I added a `System Requirements` Section in the top `Dependencies` section with Daniel's guidance on how to estimate the amount of RAM needed for various simulations.

kasparm (approved):
Looks good to me.
Thanks for your work.

README.md
```diff
@@ -1,50 +1,248 @@
...
+This package can also be run using Docker, in which the following are required:
+- [Docker]
+- [Gurobi Cloud License] file
 
 For sample data to use with the simulation, please visit [Zenodo].
 
+### System Requirements
+
+Large simulations (e.g. covering the entire USA grid instead of specific interconnects or with large time intervals) can constrained by the amount of RAM. As a general estimate, 1-2 GB of RAM is needed per hour in the interval in a simulation across the entire USA grid. For example, a 24-hour interval would require 24-48 GB of RAM; if only 16 GB of RAM is available, consider using a time interval of 8 hours or less. The memory necessary would also be proportional to the size of grid used, so as the Western interconnect is ~8 times smaller than the entire USA grid, it would also take 8x less RAM.
+
```
rouille, line 31: There are a couple of sentences above that sound weird but I am not a native speaker so I am likely wrong:
* can constrained by the amount of RAM --> can require a significant amount of RAM
* As a general estimate, 1-2 GB of RAM is needed per hour in the interval --> As a general estimate, 1-2 GB of RAM is needed to solve the DCOPF problem in the full U.S. grid over 1-hour.

rouille (approved):
Looks great. Thanks.

danielolsen (approved):
Thanks.

PR #95, opened 2020-12-22
REST API to launch simulation

jon-hagg:
### Purpose
Enable powersimdata to communicate with the reisejl container over http, which along with using a shared volume will eliminate the need for ssh within a container based installation. 

### What it does
Exposes an api endpoint as an alternative way to invoke `call.py` . This can be called by powersimdata using only the hostname of the container and the scenario id (see the sample curl command in `app.py`). Note, we could optionally use this on the server but that's not the primary goal here. The `state.py` module has some helper classes to keep track of which scenarios have been started during the lifetime of the flask app and provide a convenient way to query them (returning the status plus stdout/stderr). I also added a "failed" status for the execute list if any exception occurs during the process. 

Minor details - the docker image will now run flask by default so I updated the readme command to override that with bash. Will add more details to the readme once the necessary support is added to powersimdata. For now, this can be run using either `flask run` on the command line or `docker run -p 5000:5000 reisejl` after building the image.

### Testing
There are some unit tests for the state classes and I've done some manual testing by setting the values in `const.py` to my local directory and mocking out the `main` method with test output. 

Update: I've tested this locally in a container by running a simulation with the cloud license and extracting the results

### Time to review
60 mins - guessing this will be new for a good chuck of people, so I can provide more details or talk through any parts that are of interest

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          python-version: [3.6, 3.7, 3.8]
+          python-version: [3.7, 3.8]
 
     name: Python ${{ matrix.python-version }}
     steps:
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN wget -q https://julialang-s3.julialang.org/bin/linux/x64/1.5/julia-1.5.3-lin
 
 WORKDIR /build/gurobi_installer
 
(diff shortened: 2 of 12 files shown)


rouille:
It looks good. How do you want to coordinate with @ahurli since you both modify the README in your PRs?

jon-hagg:
Either way, if the documentation PR is merged first I will rebase and fix any conflicts.

rouille (approved):
Thanks

PR #103, opened 2021-01-29
feat: implement more storage features

danielolsen:
### Purpose

Interpret and use all of the storage features that we specify.

### What is the code doing

- We include self-discharge losses on the state of charge across all hours
- We constrain the ending state of charge

### Testing

The `LossFactor` parameter has been tested in scenario 2129, which is identical to scenario 2124 besides adding `Lossfactor = 0.05` for each storage device. 100% efficiency in both cases
- Scenario 2124: 10,755,668 MWh charged, 10,769,628 MWh discharged. (making use of a small amount of 'free' energy at the start of the first interval). Storage is sometimes 'held' for a while.
image

- Scenario 2129: 11,879,526 MWh charged, 8,883,825 MWh discharged. Energy is discharged very soon after it is charged, it is not typically 'held' because this is wasteful.
image

The terminal storage parameters were tested in scenario 2128, but with the wrong sign on the TerminalMax constraint, so it was constrained to be 60%-100% at the end of each interval, rather than 40-60%. This can be confirmed by checking `storage_e[storage_e.index.hour == 23].min()` and `storage_e[storage_e.index.hour == 23].max()`, and comparing to scenario 2124 (the same, except for these terminal storage constraints). This is being re-tested in scenario 2134.

### Time to review

15 minutes.

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -275,7 +275,7 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
         JuMP.@constraint(m,
             soc_tracking[i in 1:sets.num_storage, h in 1:(num_hour-1)],
             storage_soc[i, h+1] == (
-                storage_soc[i, h]
+                storage_soc[i, h] * (1 - storage.sd_table.LossFactor[i])
                 + storage.sd_table.InEff[i] * storage_chg[i, h+1]
                 - (1 / storage.sd_table.OutEff[i]) * storage_dis[i, h+1]),
             container=Array)
--- a/src/model.jl
+++ b/src/model.jl
@@ -287,6 +287,16 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
                 + storage.sd_table.InEff[i] * storage_chg[i, 1]
                 - (1 / storage.sd_table.OutEff[i]) * storage_dis[i, 1]),
             container=Array)
+        println("storage final_soc_min: ", Dates.now())
+        JuMP.@constraint(m,
+            soc_terminal_min[i in 1:sets.num_storage],
+            storage_soc[i, num_hour] >= storage.sd_table.ExpectedTerminalStorageMin[i],
+            container=Array)
+        println("storage final_soc_max: ", Dates.now())
+        JuMP.@constraint(m,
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
Looks good.

lanesmith218 (approved):
Looks good to me too! I have a couple questions, but they don't prevent this code from being merged. The optimization is daily with perfect foresight, right? Also, you mention constraining the SOC to be between 40% and 60% at the end of each interval. Why are those bounds chosen? I like the option for allowing final SOC to not be forced to equal beginning SOC, as these models often seem to do.

danielolsen:
@lanesmith218
- We do run with perfect foresight, but we can choose the length of each interval. 24h is the default.
- 40-60% was arbitrarily input in the scenario setup process, just to demonstrate the feature.
- By not forcing ending SoC to be equal to starting SoC, we give the optimizer more flexibility. Any different in energy at the end of the period is not 'free', since the objective function is penalized for ending the interval with less energy than the start. I think cyclical constraints could make sense if we are running many individual days in parallel, but make less sense in a serial simulation like this.

lanesmith218:
> @lanesmith218
> 
> * We do run with perfect foresight, but we can choose the length of each interval. 24h is the default.
> * 40-60% was arbitrarily input in the scenario setup process, just to demonstrate the feature.
> * By not forcing ending SoC to be equal to starting SoC, we give the optimizer more flexibility. Any difference in energy at the end of the period is not 'free', since the objective function is penalized for ending the interval with less energy than the start. I think cyclical constraints could make sense if we are running many individual days in parallel, but make less sense in a serial simulation like this.

Great, thank you for your response!

danielolsen:
Scenario 2134 behaves as expected with the fixed terminal storage soc constraints:
```python
>>> storage_e = scenario.state.get_storage_e()
>>> storage_e[storage_e.index.hour == 23].min()
0    1674.034668
1    1674.034668
2    1674.034668
3    1674.034668
4    1674.034668
5    1674.034668
dtype: float32
>>> storage_e[storage_e.index.hour == 23].max()
0    2511.052002
1    2511.052002
2    2511.052002
3    2511.052002
4    2511.052002
5    2511.052002
dtype: float32
```

PR #105, opened 2021-02-02
Non blocking reads from process output

jon-hagg:
### Purpose
The api call that launches a simulation works by creating a process with `Popen` and returning a json response that includes stdout/stderr. This content is updated each time the status endpoint is called by reading from the standard streams. Problem is the current way it's done blocks until the process is complete, which is typically a long time and not a good experience.

### What it does
Makes it so the http call returns immediately, by moving the blocking call to a separate thread, and using a thread safe queue to buffer output between that and the request thread. This way any call to either launch or check on a simulation will return the stdout/stderr even if the simulation is still running. 

### Testing
I had to fix the tests after refactoring, which ended up being good since now we are not using any fakes, so the tests are running the actual code. Since concurrency is involved, it's not totally clear how reliable this is, but the tests are simple enough that it seems to be stable.

### Time to review
10-30 mins

### DIFF
--- a/pyreisejl/utility/state.py
+++ b/pyreisejl/utility/state.py
@@ -21,8 +21,10 @@ class Listener:
         t.start()
 
     def _enqueue_output(self):
-        for line in iter(self.stream.readline, b""):
-            self.queue.put(line)
+        for line in self.stream:
+            s = line.decode().strip()
+            if len(s) > 0:
+                self.queue.put(s)
         self.stream.close()
 
     def poll(self):
@@ -34,7 +36,7 @@ class Listener:
         try:
             while True:
                 line = self.queue.get_nowait()
-                result.append(line.decode().strip())
+                result.append(line)
         except Empty:  # noqa
             pass
         return result
--- a/pyreisejl/utility/state.py
+++ b/pyreisejl/utility/state.py
@@ -1,24 +1,57 @@
 from dataclasses import dataclass, field
-from typing import Any, Dict
+from queue import Empty, Queue
+from threading import Thread
+from typing import Any, Dict, List
 
 from pyreisejl.utility.helpers import get_scenario_status
 
 
+def _enqueue_output(stream, queue):
+    for line in iter(stream.readline, b""):
+        queue.put(line)
+    stream.close()
+
+
+def _start_thread(stream, queue):
+    t = Thread(target=_enqueue_output, args=(stream, queue))
+    t.daemon = True
+    t.start()
+
+
+def _read_stream(q):
+    result = []
+    try:
+        while True:
+            line = q.get_nowait()
+            result.append(line.decode().strip())
+    except Empty:  # noqa
+        pass
+    return result
+
+
 @dataclass
 class SimulationState:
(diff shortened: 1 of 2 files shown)


pyreisejl/utility/state.py
```diff
@@ -1,24 +1,68 @@
...
+        self.queue = Queue()
+        self._start()
+
+    def _start(self):
+        t = Thread(target=self._enqueue_output)
+        t.daemon = True
+        t.start()
+
+    def _enqueue_output(self):
+        for line in iter(self.stream.readline, b""):
```
ToddG, line 24: https://docs.python.org/3/library/io.html
"""Note that it’s already possible to iterate on file objects using for line in file: ... without calling file.readlines()."""

From this, could you simplify this?

"""for line in self.stream:"""

pyreisejl/utility/state.py
```diff
@@ -1,24 +1,68 @@
...
 from pyreisejl.utility.helpers import get_scenario_status
 
 
+class Listener:
+    """Runs in the background to read from stdout/stderr of a long lived
+    process"""
+
+    def __init__(self, stream):
+        self.stream = stream
+        self.queue = Queue()
```
ToddG, line 15: Unbounded Queue leaves you vulnerable to OOM issues. Would it be better to limit and throw an exception when out of bounds, or shed oldest entries?

ToddG (approved):
Looks like a great use case. I have just a few minor questions. Feel free to address or not.

pyreisejl/utility/state.py
jon-hagg: Yeah that is better, think I will do that

pyreisejl/utility/state.py
```diff
@@ -1,24 +1,68 @@
...
+
+    def poll(self):
+        """Get the latest output from the stream
+
+        :return: (*list*) -- list of lines since previous poll
+        """
+        result = []
+        try:
+            while True:
+                line = self.queue.get_nowait()
```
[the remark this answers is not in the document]
jon-hagg: Just making sure, does the `except` block below already do this?

pyreisejl/utility/state.py
```diff
@@ -1,24 +1,68 @@
...
+        self._start()
+
+    def _start(self):
+        t = Thread(target=self._enqueue_output)
+        t.daemon = True
+        t.start()
+
+    def _enqueue_output(self):
+        for line in iter(self.stream.readline, b""):
+            self.queue.put(line)
```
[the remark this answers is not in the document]
jon-hagg: Hasn't been an issue, but I would want to filter those

pyreisejl/utility/state.py
jon-hagg: Good point, not sure how much of an issue it is but I will make a note to revisit this. For context, the size of the streams is much smaller than the output data, which is most likely to cause OOM issues.

pyreisejl/utility/state.py
jon-hagg: Done

PR #106, opened 2021-02-02
refactor: remove Gurobi requirement

danielolsen:
### Purpose

Removes the requirement that Gurobi.jl be installed as a requirement of REISE.jl, since it can now be run with any optimizer (Closes #101). Allows the user to pass solver settings from python (Closes #102).

### What is the code doing

- In **src/solver_specific/gurobi.jl**: we move all code that depends on importing Gurobi (from **REISE.jl** and **loop.jl**)
  - `run_scenario_gurobi` wraps around `run_scenario` as defined in **src/REISE.jl**.
  - The new function definition with the call signature `new_model(env::Gurobi.Env)` creates another method for the `new_model` function defined in **loop.jl**; the one in **loop.jl** will take any type, but if we pass a `Gurobi.Env` object then Julia will use the most specific method available, the one defined in **gurobi.jl** (see https://docs.julialang.org/en/v1/manual/methods/ for more details).
- In **src/REISE.jl**:
  - We remove the `Gurobi` import, instead importing `Requires`, and using this package to only include **src/solver_specific/gurobi.jl** if Gurobi is also imported.
  - We allow a dictionary `solver_kwargs` to be passed to `run_scenario`.
  - We return the `JuMP.Model` instead that we built in `interval_loop` so that when we are running with the shared gurobi environment in we can free the Gurobi environment in the most proper way according to their documentation.
- In **src/loop.jl**, in `interval_loop`:
  - We refactor how we check which solver is being used to a more general form.
  - We return the `JuMP.Model` instance, to get returned via `run_scenario` and cleaned up as necessary in `run_scenario_gurobi`
- In **pyreisejl/utility/call.py**:
  - We refactor the `launch_scenario` function into a class with methods, and we define a gurobi-specific subclass that does gurobi-specific things.
  - We refactor `main` to use this new object-based launch method.

#### Julia

From Julia, we can verify that this works by creating fresh environments, installing REISE.jl along with with different solvers (note that Gurobi.jl is not installed by default), and running as desired:

##### With Gurobi.jl

```julia
julia> import Pkg

julia> Pkg.activate("Gurobitest")
 Activating new environment at `C:\Users\DanielOlsen\repos\bes\Gurobitest\Project.toml`

julia> Pkg.develop(path=raw"C:\Users\DanielOlsen\repos\bes\REISE.jl")
Path `C:\Users\DanielOlsen\repos\bes\REISE.jl` exists and looks like the correct package. Using existing path.
  Resolving package versions...
[...]

PR #108, opened 2021-02-09
fix: add parentheses for proper boolean algebra

danielolsen:
### Purpose

Fix a bug introduced in https://github.com/Breakthrough-Energy/REISE.jl/pull/106.

### Details

Several scenarios have failed on the server. I managed to get a traceback (requires running from julia manually on the server, since these Exceptions don't make it into the recorded stdout/stderr in our current setup) and what I got is:
```julia
ERROR: MethodError: no method matching &(::Bool, ::String)
Closest candidates are:
  &(::Any, ::Any, ::Any, ::Any...) at operators.jl:538
  &(::Bool, ::Missing) at missing.jl:160
  &(::Bool, ::Bool) at bool.jl:40
  ...
Stacktrace:
 [1] interval_loop(::Gurobi.Env, ::Dict{String,Any}, ::Dict{String,Int64}, ::Int64, ::Int64, ::Int64, ::String, ::String, ::Int64) at /home/bes/pcm/REISE.jl/src/loop.jl:138
 [2] (::REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}})() at /home/bes/pcm/REISE.jl/src/REISE.jl:85
 [3] #83 at /home/bes/pcm/REISE.jl/src/save.jl:113 [inlined]
 [4] redirect_stderr(::REISE.var"#83#87"{REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}}}, ::IOStream) at ./stream.jl:1150
 [5] #82 at /home/bes/pcm/REISE.jl/src/save.jl:112 [inlined]
 [6] redirect_stdout(::REISE.var"#82#86"{IOStream,REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}}}, ::IOStream) at ./stream.jl:1150
 [7] #81 at /home/bes/pcm/REISE.jl/src/save.jl:111 [inlined]
 [8] open(::REISE.var"#81#85"{IOStream,REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}}}, ::String, ::Vararg{String,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./io.jl:325
 [9] open at ./io.jl:323 [inlined]
 [10] #80 at /home/bes/pcm/REISE.jl/src/save.jl:110 [inlined]
 [11] open(::REISE.var"#80#84"{REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}},String}, ::String, ::Vararg{String,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./io.jl:325
 [12] open at ./io.jl:323 [inlined]
 [13] redirect_stdout_stderr(::REISE.var"#95#96"{Int64,String,Gurobi.Env,Int64,Dict{String,Any}}, ::String, ::String) at /home/bes/pcm/REISE.jl/src/save.jl:109
 [14] run_scenario(; num_segments::Int64, interval::Int64, n_interval::Int64, start_index::Int64, inputfolder::String, outputfolder::Nothing, threads::Int64, optimizer_factory::Gurobi.Env, solver_kwargs::Dict{String,Int64}, resume_at::Int64) at /home/bes/pcm/REISE.jl/src/REISE.jl:83
 [15] run_scenario_gurobi(; solver_kwargs::Nothing, kwargs::Base.Iterators.Pairs{Symbol,Any,NTuple{5,Symbol},NamedTuple{(:interval, :n_interval, :start_index, :inputfolder, :threads),Tuple{Int64,Int64,Int64,String,Int64}}}) at /home/bes/pcm/REISE.jl/src/solver_specific/gurobi.jl:9
 [16] top-level scope at REPL[3]:1
```

From looking at what is recorded in stdout, we see that the last interval encountered a non-optimal result (I've seen both Numeric and Suboptimal), and that neither load shedding nor the `BarHomogeneous` setting of Gurobi were enabled, so based on the logic in `interval_loop` we should be modifying the problem and re-running if not for this Exception.

Looking at the [Julia Operator Precedence](https://docs.julialang.org/en/v1/manual/mathematical-operations/#Operator-Precedence-and-Associativity), this error makes sense, since `&` has precedence over `==`, but there's no built-in way to `&` a `Bool` and a `String`.

### What is the code doing

Adding parentheses, so that the equality is evaluated to a `Bool` before any other binary logic operations.

### Testing

Approximately 40 scenarios failed with the old code, re-running the new code they all completed successfully.

### Time to review

5 minutes.

### DIFF
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -135,7 +135,7 @@ function interval_loop(factory_like, model_kwargs::Dict,
                 results = get_results(f, voi, model_kwargs["case"])
                 break
             elseif ((status in numeric_statuses)
-                    & JuMP.solver_name(m) == "Gurobi"
+                    & (JuMP.solver_name(m) == "Gurobi")
                     & !("BarHomogeneous" in keys(solver_kwargs)))
                 # if Gurobi, and BarHomogeneous is not enabled, enable it and re-solve
                 solver_kwargs["BarHomogeneous"] = 1
@@ -150,7 +150,7 @@ function interval_loop(factory_like, model_kwargs::Dict,
                 JuMP.set_optimizer_attributes(m, pairs(solver_kwargs)...)
                 m, voi = _build_model(m; symbolize(model_kwargs)...)
                 intervals_without_loadshed = 0
-            elseif (JuMP.solver_name(m) == "Gurobi"
+            elseif ((JuMP.solver_name(m) == "Gurobi")
                     & !("BarHomogeneous" in keys(solver_kwargs)))
                 # if Gurobi, and BarHomogeneous is not enabled, enable it and re-solve
                 solver_kwargs["BarHomogeneous"] = 1


BainanXia (approved):
Make sense. Previously, I was thinking in the Python way, in which A and B == C, B == C has priorities.

PR #112, opened 2021-02-18
Define glpk launcher and expose command line and api option

jon-hagg:
### Purpose
Add new `--solver` parameter (and corresponding query string) which can currently be either `gurobi` or `glpk`. This enables running a simulation in test scenarios where a license is not available. 

### What it does
* implement required launcher subclass
* install glpk in docker image
* make the launcher class determined (optionally) by end user, defaulting to gurobi
* move launchers to separate file, to keep `call.py` relatively clean

### Time to review
10 min

### DIFF
--- a/pyreisejl/utility/call.py
+++ b/pyreisejl/utility/call.py
@@ -27,6 +27,18 @@ def _record_scenario(scenario_id, runtime):
     )
 
 
+def _get_launcher(solver):
+    """Determine the launcher type given value from command line
+
+    :param str solver: user provided solver name
+    :return: (*type*) -- the launcher type, which can be instantiated
+    """
+    launch_map = {"gurobi": GurobiLauncher, "glpk": GLPKLauncher}
+    if solver is None:
+        return GurobiLauncher
+    return launch_map[solver]
+
+
 def main(args):
     # Get scenario info if using PowerSimData
     if args.scenario_id:
@@ -50,14 +62,7 @@ def main(args):
         )
         raise WrongNumberOfArguments(err_str)
 
-    launch_map = {"gurobi": GurobiLauncher, "glpk": GLPKLauncher}
-    solver = args.solver
-    if solver:
-        _launcher = launch_map[solver]
-    else:
-        _launcher = GLPKLauncher
-
-    launcher = _launcher(
+    launcher = _get_launcher(args.solver)(
         args.start_date,
         args.end_date,
         args.interval,
--- a/pyreisejl/utility/call.py
+++ b/pyreisejl/utility/call.py
@@ -8,7 +8,7 @@ from pyreisejl.utility.helpers import (
     insert_in_file,
     sec2hms,
 )
-from pyreisejl.utility.launchers import GLPKLauncher, GurobiLauncher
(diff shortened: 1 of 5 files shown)


pyreisejl/utility/call.py
```diff
@@ -34,109 +27,16 @@ def _record_scenario(scenario_id, runtime):
...
-
-        runtime = round(end - start)
-        hours, minutes, seconds = sec2hms(runtime)
-        print(f"Run time: {hours}:{minutes:02d}:{seconds:02d}")
-
-        return runtime
+    launch_map = {"gurobi": GurobiLauncher, "glpk": GLPKLauncher}
+    if solver is None:
+        return GurobiLauncher
+    return launch_map[solver]
```
danielolsen, line 39: Should we provide a meaningful exception in case the user specifies something besides the keys in `launch_map`?

Should we allow a little more flexibility so that strings like `"Gurobi"` or `"GLPK"` still resolve successfully?

Should we define `launch_map` within the body of **launchers.py** so that we can do a single import?

pyreisejl/utility/parser.py
```diff
@@ -73,6 +73,10 @@ def parse_call_args():
         "instead of deleted after extraction. "
         "This flag is only used if the extract-data flag is set.",
     )
+    parser.add_argument(
+        "--solver",
+        help="Specify the solver to run the optimization. Will default to gurobi",
+    )
```
danielolsen, line 79: If we define `launch_map` within **launchers.py**, does that let us do something like automatically populate the help string with a list of valid keys?

pyreisejl/utility/call.py
jon-hagg: Affirmative, all good ideas. I probably could have made this a draft PR since it's a bit rough currently.

danielolsen: I like where it's headed so far.

pyreisejl/utility/launchers.py
```diff
@@ -0,0 +1,172 @@
...
+            outputfolder=self.execute_dir,
+            optimizer_factory=GLPK.Optimizer,
+        )
+        end = time()
+
+        runtime = round(end - start)
+        hours, minutes, seconds = sec2hms(runtime)
+        print(f"Run time: {hours}:{minutes:02d}:{seconds:02d}")
+
+        return runtime
```
danielolsen, line 119: If we want to maximize our DRY, we could define a `round_and_print_runtime` method of the `Launcher` class and then call `return self.round_and_print_runtime(start, end)`, but it's not really necessary.

danielolsen:
The code looks good. Has it been tested?

jon-hagg:
> The code looks good. Has it been tested?

Yeah, just manual tests in a container - checked with invalid solver and completed a simulation with glpk

pyreisejl/utility/launchers.py
jon-hagg: Yeah I wanted to reuse the code here but since it's interleaved it felt like the result would be harder to read. If we end up adding more solvers I would want to do something about it.

danielolsen (approved):
Looks good, thanks!

PR #116, opened 2021-03-03
Improve docker build caching

jon-hagg:
### Purpose
Reduce expected build time by making use of docker caching. We do this by copying just the requirements.txt and toml files earlier in the dockerfile since they change less often, and the bulk of the time is installing julia packages. Adding the code directories later means changes to the code won't invalidate package caches, and we won't need to reinstall packages if they didn't change. For context, each instruction in the dockerfile represents a "layer" of the resulting filesystem, and each layer can be cached, but changes to one layer invalidates the cache for subsequent instructions.

### What it does
Rearranged the dockerfile. I'm pretty sure there is a better way to do this but it works for now.

### Testing
Ran a simulation using both gurobi and glpk with the resulting image. 

### Time to review
5-10 mins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,8 +19,10 @@ ENV PATH="$PATH:/usr/share/julia-1.5.3/bin" \
     FLASK_APP=pyreisejl/utility/app.py
 
 WORKDIR /app
-COPY . .
+COPY requirements.txt .
+RUN pip install -U pip; pip install -r requirements.txt
 
+COPY . .
 RUN julia -e 'using Pkg; \
 	Pkg.activate("."); \
 	Pkg.instantiate(); \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -19,17 +19,27 @@ ENV PATH="$PATH:/usr/share/julia-1.5.3/bin" \
...
 RUN julia -e 'using Pkg; \
 	Pkg.activate("."); \
 	Pkg.instantiate(); \
 	Pkg.add("Gurobi"); \
+	Pkg.add("GLPK")'
+
+COPY src src
+
+RUN julia -e 'using Pkg; \
+	Pkg.activate("."); \
```
rouille, line 38: Do we need to invoke `Pkg.activate(".")` twice?

jon-hagg: Nope, thanks.

rouille (approved):
Looks good

PR #118, opened 2021-03-31
Incorporate a model of demand flexibility

lanesmith:
### Purpose

The purpose of this new code is to incorporate a model of demand flexibility into the production cost model. This is a first iteration of the model and will be updated in the future to (1) become more computationally efficient and (2) incorporate the modeling efforts of our external collaborators working on sectoral electrification.

### What is the code doing?

The new code includes a model of demand flexibility and supporting functionality to prepare the relevant flexibility data and process the new results. Here is a more in-depth breakdown of the additions by file:
- `src/REISE.jl`: brief updates to the overall framework to allow demand flexibility to be considered.
- `src/loop.jl`: very brief update to include the `Flexibility` object. While not needed in this iteration of the code, information from the `Flexibility` object will almost certainly be needed once we start considering the look-ahead period.
- `src/model.jl`: this contains the bulk of the code additions. The `JuMP` model is updated to include new decision variables and constraints that are needed for the demand flexibility model. Some preexisting constraints needed to be updated following the inclusion of the demand flexibility model. Functionality is included to map the demand flexibility profile data (which is originally by load zone) to each bus that contains demand flexibility (which is currently every bus with demand).
- `src/types.jl`: create a new `Flexibility` object to store relevant demand flexibility information. Update preexisting types as needed.
- `src/read.jl`: provide functionality to read in a single demand flexibility profile and other associated demand flexibility parameters (e.g., shifting duration). The original plan was to allow for multiple profiles to be read in, thereby allowing the different demand sectors to remain disaggregated. However, after internal discussions, it seems like it is probably best to aggregate these profiles; this functionality would probably be best suited for `PreREISE` (similar to what is done in `PreREISE/prereise/gather/demanddata/nrel_efs/aggregate_demand.py`). Right now, I'm just reading in the demand flexibility parameters (currently just duration, though curtailment cost could be included at a later time) from a very basic .csv file. I'm open to suggestions on how to best store and read in this data (though this design may remain pretty fluid depending whether the duration parameter is static or more dynamic depending on weather).
- `src/query.jl`: accesses the `load_shift_dn` and `load_shift_up` variables from the `VariablesOfInterest` object.
- `src/save.jl`: saves the `load_shift_dn` and `load_shift_up` variables.
- `pyreisejl/utility/extract_data.py`: allows the `load_shift_dn` and `load_shift_up` variables to be accessed and saved as .pkl files.
- `README.md`: update the documentation to include new the demand flexibility model.

### Where to look?

Most of the new code is located within the `src` folder to update the model (`src/REISE.jl`, `src/loop.jl`, and `src/model.jl`) and the supporting functions to prepare the data (`src/read.jl` and `src/types.jl`) and process the results (`src/query.jl` and `src/save.jl`). I also update `pyreisejl/utility/extract_data.py` to allow the Python wrapper functions to work with the new model. Lastly, I also updated `README.md` to appropriately document the new model.

### Testing

The best that I can tell, none of the preexisting tests need to be modified following the inclusion of this code. Additionally, none of the code that I added appears to warrant its own tests (based on current testing convention within this repository). I did run some simple tests on my local machine to serve as a proof of concept. Essentially, the demand flexibility model behaved in a way that was to be expected, with demand trying to shift from times of higher system prices (which can be visualized via the LMP representation) to times of lower system prices. Predictably, the inclusion of the demand flexibility model (in its current state) really slows down the solution time, even when using Gurobi. While I don't expect the difference to be as stark when executed on the server, these results indicate the need for some form of feature reduction, which will be incorporated in the future. For a little more information on these results, please refer to my recent presentation, which is located in the `Deliverables` folder of my `Explorations` folder.

### Visuals

While I don't include any visuals here, there are a couple limited visuals in the presentation that I mentioned in the previous section. That presentation is located in the `Deliverables` folder of my `Explorations` folder.

### Time estimate

I'm not sure. I tried to submit this PR a little earlier, so there should be fewer lines of code to review than in the past. However, it does incorporate a new model and works across quite a few different files in `REISE.jl`.


README.md
```diff
@@ -619,6 +627,11 @@ Power available from time-varying generator (hydro, wind, solar) ![i](https://re
 at time ![t](https://render.githubusercontent.com/render/math?math=t).
 - ![x_{l}](https://render.githubusercontent.com/render/math?math=x_%7Bl%7D):
 Impedance of branch ![l](https://render.githubusercontent.com/render/math?math=l).
+- ![\overline{\delta}_{b, t}](https://render.githubusercontent.com/render/math?math=%5Coverline%7B%5Cdelta%7D_%7Bb%2Ct%7D):
+Demand flexibility available at bus ![b](https://render.githubusercontent.com/render/math?math=b)
+at time ![t](https://render.githubusercontent.com/render/math?math=t).
```
danielolsen, line 632: Are there different limits for each direction?

lanesmith: Not right now. I think they should probably be different though. If this were a single consumer model, I would probably set the upper bound on `load_shift_up` to be something like the consumer's maximum demand. However, since we are dealing with demand from different sectors, where demand can vary a lot (e.g., based on weather, based on EV travel), I don't feel like the maximum nodal demand is a good metric. My feeling is that the upper bound on `load_shift_up` should be a value in between the current upper bound (i.e., the value obtained from the demand flexibility profile) and the maximum nodal demand (which I know is quite a large spread). This is something that warrants a little further exploration.

pyreisejl/utility/extract_data.py
```diff
@@ -250,6 +256,12 @@ def _get_outputs_from_converted(matfile):
     except AttributeError:
         pass
 
+    try:
```
danielolsen, line 259: Where do we expect an AttributeError to come from? For storage and DC lines, we expect it from a missing field in `mpc` (if no storage or no DC lines), but `mpc` will always have `bus`.

src/model.jl
```diff
@@ -227,8 +278,7 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
     end)
     if load_shed_enabled
         JuMP.@variable(m,
-            0 <= load_shed[i in 1:sets.num_load_bus, j in 1:interval_length]
-            <= bus_demand[sets.load_bus_idx[i], j],
+            load_shed[i in 1:sets.num_load_bus, j in 1:interval_length],
```
danielolsen, line 281: It seems like with this change we can shed more load than we have available at a bus?

src/model.jl
```diff
@@ -258,18 +323,40 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
...
+            )
+        else
+            JuMP.@constraint(
+                m, 
+                load_shed_ub[i in 1:sets.num_load_bus, j in 1:interval_length], 
+                0 <= load_shed[i, j] <= bus_demand[sets.load_bus_idx[i], j]
+            )
+        end
+    end
+
```
danielolsen, line 359: Oh I see, load shedding is constrained here now. We will need to change how we update the model during the different intervals.

danielolsen:
What do the values in the new **flexibility.csv** input represent? Are they fractions of the demand that are flexible over the course of the year?

src/model.jl
```diff
@@ -258,18 +323,40 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
     if load_shed_enabled
         injections = JuMP.@expression(m, injections + load_bus_map * load_shed)
     end
-    withdrawls = JuMP.@expression(m, bus_demand)
+    withdrawals = JuMP.@expression(m, bus_demand)
```
danielolsen, line 326: Good catch

pyreisejl/utility/extract_data.py
lanesmith: Great point. When I was first figuring out how demand flexibility would be included, I anticipated having a set that identified demand flexibility nodes (similar to storage). However, I saw what was done with the `bus` field and ended up going with that instead. Even once feature reduction is implemented, I can still probably set the output IDs using the `bus` field, so something like this for demand flexibility won't likely be needed in the future either. I will remove this.

src/model.jl
lanesmith: I ended up constraining the `load_shed` variable separate from the variable creation. When demand flexibility is enabled, `load_shift_up` and `load_shift_dn` will have an effect on the upper bound of `load_shed`. I wasn't sure how flexible setting constraints within the variable creation was, so that's why I moved it to its own constraint.

danielolsen: Here is how we modify variable bounds as we step through intervals: https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/loop.jl#L86

lanesmith: Thanks for pointing this out! I was just going to reply to your next comment suggesting that this is one of changes that I forgot to account for. Is it preferred to keep the `load_shed` variable the same as it was originally (include the constraints in the variable creation) and instead account for `load_shift_up` and `load_shift_dn` in `loop.jl` if demand flexibility is allowed?

lanesmith:
> What do the values in the new **flexibility.csv** input represent? Are they fractions of the demand that are flexible over the course of the year?

They are an amount of demand (in MW) that is considered to be flexible at that time. So, if you divide the amount of flexible demand provided by **flexibility.csv** by the total demand provided by **demand.csv**, you would get the percentage of demand at that bus that is flexible.

src/model.jl
danielolsen: We need to make sure that load shedding is appropriately constrained in the first iteration of the loop (when we run `_build_model`), as well as appropriately updated in each iteration. We will have to do the same for flexibility (if present).

BainanXia: @danielolsen I'm about to point the same thing out.

pyreisejl/utility/extract_data.py
lanesmith: Done.

src/model.jl
lanesmith: I have updated `loop.jl`, so `load_shed` and the demand flexibility variables (`load_shift_up` and `load_shift_dn`) should be properly constrained in each iteration.

README.md
BainanXia: Then should we use two different parameters to maximize the flexibility for the sake of future usage and define them to be equal at this point. It could be simply noted down here in the readme without modifying the code for now to avoid complaints of redundant parameters (?).

lanesmith: Yeah, good idea. I will update the documentation

lanesmith: Done.

README.md
```diff
@@ -619,6 +627,14 @@ Power available from time-varying generator (hydro, wind, solar) ![i](https://re
...
 - ![x_{l}](https://render.githubusercontent.com/render/math?math=x_%7Bl%7D):
 Impedance of branch ![l](https://render.githubusercontent.com/render/math?math=l).
+- ![\underline{\delta}_{b, t}](https://render.githubusercontent.com/render/math?math=%5Cunderline%7B%5Cdelta%7D_%7Bb%2Ct%7D):
+Demand flexibility curtailments available at bus ![b](https://render.githubusercontent.com/render/math?math=b)
+at time ![t](https://render.githubusercontent.com/render/math?math=t).
+- ![\overline{\delta}_{b, t}](https://render.githubusercontent.com/render/math?math=%5Coverline%7B%5Cdelta%7D_%7Bb%2Ct%7D):
+Demand flexibility additions available at bus ![b](https://render.githubusercontent.com/render/math?math=b)
+at time ![t](https://render.githubusercontent.com/render/math?math=t).
+- ![\Delta](https://render.githubusercontent.com/render/math?math=%7B%5CDelta%7D):
+The length of the rolling load balance window, used to account for the duration that flexible demand is deviating from the base demand.
```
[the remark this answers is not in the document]
lanesmith: Good points. I'm open to other variable name suggestions if there is something that you think fits particularly well. I initially settled on `Delta` because it aligned with the similar `deltas` I was using for demand shifting. As for this parameters becoming a variable, I certainly think that's something we can explore in the future. If it's alright with you, I think we should keep it as a parameter for now, especially while we're trying to integrate other flexibility profiles and aggregate those profiles by sector.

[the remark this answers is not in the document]
danielolsen: Other `deltas` are `MW`, this `Delta` is hours?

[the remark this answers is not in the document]
lanesmith: Correct. I'll be sure to specify those units. "Aligned" probably wasn't the right word above. I wasn't trying to say that these variables/parameters are similar, rather that they pertain to a similar part of the model (demand flexibility).

src/REISE.jl
```diff
@@ -9,9 +9,9 @@ import Requires
 import SparseArrays: sparse, SparseMatrixCSC
 
 
-include("types.jl")         # Defines Case, Results, Storage,
+include("types.jl")         # Defines Case, Results, Storage, Flexibility,
```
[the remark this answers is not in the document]
lanesmith: Yeah, I can rename that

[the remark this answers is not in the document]
BainanXia: We are kind of using 'flexibility' to refer 'demand flexibility' both in variable names and doc strings. Not sure we should be more specific just in case we might define other flexibilities in the future.

README.md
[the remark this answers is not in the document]
danielolsen: I agree with Bainan, it would be nice to be more descriptive here. Maybe `Delta_t^{balance}` or something similar?

src/REISE.jl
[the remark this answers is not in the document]
lanesmith: Yeah, that's a good point. I'll be sure change `Flexibility` to `DemandFlexibility` in `types.jl` (which is where that notation currently originates). I'll make sure other doc strings are updated similarly

src/model.jl
```diff
@@ -215,6 +256,16 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
...
     end
+    # Demand flexibility parameters (if present)
+    bus_flex_amt = _make_bus_flexibility_amount(
+        case, flexibility, start_index, end_index
+    )
+    flexibility_enabled = (bus_flex_amt != zeros(sets.num_bus, interval_length))
+    if flexibility_enabled && (
+        flexibility.duration == nothing || flexibility.duration > interval_length
+    )
+        flexibility.duration = interval_length
```
BainanXia, line 267: We only support flexibility duration (delta) to be smaller or equal to `interval_length`. This is an implicit assumption which does not reflect in the mathematical formulation. Should put down a warning here to let the user know "any flexibility duration > `interval_length` will be trim down to the `interval_length`? In the future, we may revisit this to implement the work around to carry some extra information to the next interval so that we could remove this limitation.

src/model.jl
```diff
@@ -215,6 +256,16 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
         storage_map = sparse(storage_bus_idx, sets.storage_idx, 1,
                              sets.num_bus, sets.num_storage)::SparseMatrixCSC
     end
+    # Demand flexibility parameters (if present)
+    bus_flex_amt = _make_bus_flexibility_amount(
+        case, flexibility, start_index, end_index
+    )
+    flexibility_enabled = (bus_flex_amt != zeros(sets.num_bus, interval_length))
```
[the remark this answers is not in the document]
lanesmith: Yeah I think that is a great idea. For the first `flexibility_enabled` idea, I thought it wasn't necessary because of the way the Scenario framework worked (or at least how I currently understand it). If someone was going to run comparisons, wouldn't that show up as separate scenarios? I may very well be wrong on that. Regardless, what you suggested would be great for people who don't use/aren't currently using the Scenario framework and are just using input data with `REISE.jl`. I'm guessing something like this would make sense to be included as a command line input, like what is enabled through `pyreisejl/utility/call.py`?

For the second `flexibility_enabled` idea, I think that is great too. Certainly relevant when considering the rollout/availability of demand flexibility to different parts of the country. That would probably have to be some sort of profile input, likely mapping demand flexibility availability to different load zones (or similar defined areas). I can certainly look into doing that. Would you like that included in this PR?

src/model.jl
lanesmith: Yeah, I can certainly add that warning (though, to be sure, not an error that stops the simulation, right?). The future functionality that you mention is something that I was considering too, and would likely need to be paired with the look-ahead functionality that I discussed last week (and has also been discussed for energy storage). Once the look ahead is added, we will definitely pass net demand flexibility deviations to the next period (similar to passing energy storage SOC to the next period).

README.md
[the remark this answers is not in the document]
lanesmith: Done.

src/REISE.jl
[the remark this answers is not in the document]
lanesmith: Done.

src/model.jl
lanesmith: The warning regarding demand flexibility duration with respect to interval length has been added. As we discussed, the functionality for including a look-ahead period and passing remaining net demand deviations across intervals will be implemented in a future PR.

src/model.jl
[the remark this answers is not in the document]
lanesmith: I want to follow-up on these two suggestions. For the first (including an overarching `demand_flexibility_enabled` that dictates whether or not demand flexibility is enabled, even if the demand flexibility profile is included), I've been thinking about how something like this would be implemented. There are a couple of options that I can think of: (1) including a command line input, similar to what is enabled through `pyreisejl/utility/call.py` and (2) including a `demand_flexibility_enabled` parameter in `demand_flexibility_parameters.csv`, which is a file I am currently using to specify the duration parameter and was thinking about using to specify which combination of the balancing constraints should be enforced. If we were to implement (1), we would probably want to implement this for other resources too, right? I imagine something similar could be done for energy storage. My concerns with this is that (a) there could be a lot of inputs if we do this for a lot of resources and (b) this would only work when using the Python implementation. For (2), the implementation would probably be a little clunky, but functional. Any time a user wanted to change their preference on allowing demand flexibility, they would need to open and update a .csv file (or generate a new scenario). I would love to hear your thoughts on these options (and potentially other ideas).

For the second suggestion (being able to specify demand flexibility participation by region), I was thinking that should be saved for a future PR. I am planning to create issues for the ideas that have been discussed in this PR and offline that will be saved for future PRs. This should help us keep track of some of these desired features.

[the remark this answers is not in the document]
BainanXia: Thanks @lanesmith, that's good summary of what we have been discussed last week. I think we need inputs from @rouille @danielolsen at this point.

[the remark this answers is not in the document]
lanesmith: I added `demand_flexiblity_enabled` using (1) as I suggested above (the method where the parameter is contained within `demand_flexibility_parameters.csv`); if desired, this can be changed. I also added the ability for the user to specify (again via `demand_flexibility_parameters.csv`) the combination of interval load balance and rolling load balance constraints they would like included in the model.

BainanXia:
Looking through the conversations here, I think all the comments have been addressed so far, right? @lanesmith

lanesmith:
> Looking through the conversations here, I think all the comments have been addressed so far, right? @lanesmith

Yes, I believe so.

BainanXia (approved):
Very nice. Thanks @lanesmith Let's get @danielolsen 's green light before merging.

lanesmith:
> Very nice. Thanks @lanesmith Let's get @danielolsen 's green light before merging.

Sounds good! There is definitely room for improvements (some ideas for some of these improvements have been outlined in some issues), but I think this functionality is a good first version. It should prove as a good foundation for some of the collaborations that are ready to look at demand flexibility.

README.md
```diff
@@ -1,7 +1,7 @@
 # REISE.jl
 Renewable Energy Integration Simulation Engine.
 
-This repository contains, in the **src** folder, the Julia scripts to run the power-flow study in the U.S. electric grid. The simulation engine relies on [Gurobi] as the optimization solver.
+This repository contains, in the `src` folder, the Julia scripts to run the power-flow study in the U.S. electric grid. The simulation engine relies on [Gurobi] as the optimization solver.
```
danielolsen, line 4: Unrelated to this PR: let's make sure we update this line, to make it clear that Gurobi is not required.

README.md
```diff
@@ -619,6 +627,14 @@ Power available from time-varying generator (hydro, wind, solar) ![i](https://re
 at time ![t](https://render.githubusercontent.com/render/math?math=t).
 - ![x_{l}](https://render.githubusercontent.com/render/math?math=x_%7Bl%7D):
 Impedance of branch ![l](https://render.githubusercontent.com/render/math?math=l).
+- ![\underline{\delta}_{b, t}](https://render.githubusercontent.com/render/math?math=%5Cunderline%7B%5Cdelta%7D_%7Bb%2Ct%7D):
+Demand flexibility curtailments available (in MW) at bus ![b](https://render.githubusercontent.com/render/math?math=b)
+at time ![t](https://render.githubusercontent.com/render/math?math=t).
+- ![\overline{\delta}_{b, t}](https://render.githubusercontent.com/render/math?math=%5Coverline%7B%5Cdelta%7D_%7Bb%2Ct%7D):
+Demand flexibility additions available (in MW) at bus ![b](https://render.githubusercontent.com/render/math?math=b)
+at time ![t](https://render.githubusercontent.com/render/math?math=t).
+- ![\Delta^{balance}](https://render.githubusercontent.com/render/math?math=\Delta^{balance}):
```
danielolsen, line 636: `\Delta^{balance}` -> `\Delta^{\text{balance}}`

README.md
```diff
@@ -630,12 +646,12 @@ Discharging efficiency of storage device at bus ![b](https://render.githubuserco
 All equations apply over all entries in the indexed sets unless otherwise listed.
 
 - ![0 \le g_{i,s,t} \le g_{i,s,t}^{\text{max}}](https://render.githubusercontent.com/render/math?math=0%5Cle%20g_%7Bi%2Cs%2Ct%7D%5Cle%20g%5E%7B%5Ctext%7Bmax%7D%7D_%7Bi%2Cs%2Ct%7D):
-generator segment power is non-negative and less than the segment width.
-- ![0 \le s_{b,t} \le a^{\text{shed}} \cdot d_{b,t}](https://render.githubusercontent.com/render/math?math=0%20%5Cle%20s_%7Bb%2Ct%7D%20%5Cle%20a%5E%7B%5Ctext%7Bshed%7D%7D%20%5Ccdot%20d_%7Bb%2Ct%7D):
-load shed is non-negative and less than the demand at that bus, if load shedding is enabled.
+Generator segment power is non-negative and less than the segment width.
+- ![0 \le s_{b,t} \le a^{\text{shed}} \cdot ( d_{b,t} + \delta_{b,t}^{\text{up}} - \delta_{b,t}^{\text{down}} )](https://render.githubusercontent.com/render/math?math=0%20%5Cle%20s_%7Bb%2Ct%7D%20%5Cle%20a%5E%7B%5Ctext%7Bshed%7D%7D%20%5Ccdot%20%28%20d_%7Bb%2Ct%7D%20%2B%20%5Cdelta_%7Bb%2Ct%7D%5E%7B%5Ctext%7Bup%7D%7D%20-%20%5Cdelta_%7Bb%2Ct%7D%5E%7B%5Ctext%7Bdown%7D%7D%20%29):
+Load shed is non-negative and less than the demand at that bus, if load shedding is enabled.
```
danielolsen, line 651: `less than the demand at that bus` -> `less than the demand at that bus (including the impact of demand flexibility)`

README.md
danielolsen: How about: `The simulation engine relies on the use of an external optimization solver; any solver supported by JuMP can be used, although we have some special code that makes efficient use of the capabilities of [Gurobi]`?

src/model.jl
```diff
@@ -38,31 +38,69 @@ end
...
+    zone_idx = 1:length(zone_list)
     zone_id2idx = Dict(zone_list .=> zone_idx)
     bus_df_with_zone_load = DataFrames.innerjoin(bus_df, zone_demand, on=:zone)
     bus_share = bus_df[:, :load] ./ bus_df_with_zone_load[:, :load_sum]
     bus_zone_idx = Int64[zone_id2idx[z] for z in case.bus_zone]
     zone_to_bus_shares = sparse(
-        bus_zone_idx, bus_idx, bus_share)::SparseMatrixCSC
+        bus_zone_idx, bus_idx, bus_share
+    )::SparseMatrixCSC
+    return zone_to_bus_shares
```
danielolsen, line 65: The `return` is not necessary here, Julia automatically returns the value of the last-evaluated expression.
See https://docs.julialang.org/en/v1/manual/functions/#The-return-Keyword

src/model.jl
```diff
@@ -38,31 +38,69 @@ end
...
+)::Matrix
+    # Bus weighting
+    zone_to_bus_shares = _make_bus_demand_weighting(case, start_index, end_index)
+
+    # Demand flexibility profiles
+    simulation_demand_flex_amt = Matrix(
+        demand_flexibility.flex_amt[start_index:end_index, 2:end]
+    )
+    bus_demand_flex_amt = permutedims(simulation_demand_flex_amt * zone_to_bus_shares)
+    return bus_demand_flex_amt
```
danielolsen, line 101: Can be removed as well.

src/model.jl
```diff
@@ -215,6 +258,22 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
...
+    if demand_flexibility.enabled && (
+        demand_flexibility.duration == nothing 
+            || demand_flexibility.duration > interval_length
+    )
+        if demand_flexibility.duration > interval_length
+            @warn (
+                "Demand flexibility durations greater than the interval length are set "
+                * "equal to the interval length."
+            )
+        end
```
danielolsen, line 274: Nice.

src/model.jl
BainanXia: Agree. But I will vote for keeping these `return` statements since they are more explicit and readable even for people not familiar with Julia.

danielolsen: In that case, there are a few places where we should add them for consistency. See https://github.com/Breakthrough-Energy/REISE.jl/issues/127

src/model.jl
```diff
@@ -258,18 +333,40 @@ function _build_model(m::JuMP.Model; case::Case, storage::Storage,
...
+            )
+        else
+            JuMP.@constraint(
+                m, 
+                load_shed_ub[i in 1:sets.num_load_bus, j in 1:interval_length], 
+                load_shed[i, j] <= bus_demand[sets.load_bus_idx[i], j]
+            )
+        end
+    end
+
```
danielolsen, line 369: Can we build up the right-hand-side of this constraint (based on the ordering of variables/parameters here, not based on the 'traditional' RHS definition which is parameters only) using a JuMP expression? I.e. start with the bus demand, add load shifting as necessary, and then constrain based on this expression?

Or equivalently, start with an expression containing load shedding, add load shifting as necessary, and constrain this expression by the bus demand?

lanesmith: Yeah, that's a good idea. That would definitely be cleaner.

src/query.jl
```diff
@@ -64,9 +64,30 @@ function get_results(f::Float64, voi::VariablesOfInterest, case::Case)::Results
...
+    # These variables will only be in the results if the model has flexible demand
+    # Initialize with empty arrays, to be discarded later if they stay empty
+    load_shift_up = zeros(0, 0)
+    load_shift_dn = zeros(0, 0)
+    try
+        load_shift_up_temp = JuMP.value.(voi.load_shift_up)
+        load_shift_dn_temp = JuMP.value.(voi.load_shift_dn)
+        load_bus_map = sparse(
+            sets.load_bus_idx, 1:sets.num_load_bus, 1, sets.num_bus, sets.num_load_bus
+        )
```
danielolsen, line 76: We create the `load_bus_map` both in this block, in the load_shed extraction block, and in `_build_model`. Maybe it is worth including in the `Sets` struct, so that it's only created once.

src/query.jl
```diff
@@ -64,9 +64,30 @@ function get_results(f::Float64, voi::VariablesOfInterest, case::Case)::Results
...
+        load_shift_up = load_bus_map * load_shift_up_temp
+        load_shift_dn = load_bus_map * load_shift_dn_temp
+    catch e
+        if isa(e, MethodError)
+            # Thrown when storage variables are `nothing`
```
danielolsen, line 81: `Thrown when storage variables` -> `Thrown when load shift variables`

src/save.jl
```diff
@@ -79,7 +79,8 @@ function save_results(results::Results, filename::String;
                 ),
             )),
         )
-    # For DC lines, storage power/energy, and load_shed, save only if nonempty
+    # For DC lines, storage power/energy, load_shed, and flexible demand,
+    # save only if nonempty
```
danielolsen, line 83: I think indenting the second line of this comment would be clearer to show that it's all one sentence.

src/query.jl
lanesmith: Good catch. That was sloppy of me.

src/model.jl
lanesmith: Okay, so we'll keep the `return` statements? I can add `return` statements to some of the functions that don't have them, like `_make_branch_map` and `_make_sets`.

src/query.jl
lanesmith: That's a good idea, I can add that.

src/model.jl
danielolsen: I think that would be a good idea, thanks for taking care of that.

lanesmith:
> It sounds like someone is volunteering to improve our test coverage.

Haha I'm not so sure about that.

README.md
lanesmith: Done.

README.md
lanesmith: Done.

README.md
lanesmith: Done.

src/model.jl
lanesmith: I added return statements to all functions where it made sense. I believe this should close #127

src/save.jl
lanesmith: Done.

src/query.jl
lanesmith: Done.

src/query.jl
lanesmith: Done.

src/model.jl
lanesmith: I believe I have implemented something similar to what you were describing.

lanesmith:
@danielolsen, I refactored `read_demand_flexibility` close to what we discussed on Friday. However, I wasn't able to get rid of the nested `try/catch` statements that are used for reading the demand flexibility parameters. It appears that variables created within a `try` statement aren't stored in memory and cannot be used outside of the `try` statement (unless, I'm guessing, you make it a `global` variable, which seems unnecessary for this application, or pre-allocate the memory outside of the `try` statement, which again doesn't seem necessary for this application). The nested `try/catch` statements appear to work how we would want them to, with errors raised in the inner `try` statement only being caught by the inner `catch` statement and not tripping the outer `catch` statement.

src/model.jl
```diff
@@ -34,35 +34,74 @@ function _make_branch_map(case::Case)::SparseMatrixCSC
...
+    _make_bus_demand_weighting(case)
 
-Given a Case object, build a matrix of demand by (bus, hour) for this interval.
+Given a Case object, build a sparse matrix that indicates the weighting of each bus in 
+    each zone.
 """
-function _make_bus_demand(case::Case, start_index::Int, end_index::Int)::Matrix
+function _make_bus_demand_weighting(
+    case::Case, start_index::Int, end_index::Int
+)::SparseMatrixCSC
```
danielolsen, line 49: This function does not appear to actually need the `start_index` or `end_index` parameters. These are only used in `_make_bus_demand` to trim the entire profile duration down to the simulation window.

danielolsen:
> @danielolsen, I refactored `read_demand_flexibility` close to what we discussed on Friday. However, I wasn't able to get rid of the nested `try/catch` statements that are used for reading the demand flexibility parameters. It appears that variables created within a `try` statement aren't stored in memory and cannot be used outside of the `try` statement (unless, I'm guessing, you make it a `global` variable, which seems unnecessary for this application, or pre-allocate the memory outside of the `try` statement, which again doesn't seem necessary for this application). The nested `try/catch` statements appear to work how we would want them to, with errors raised in the inner `try` statement only being caught by the inner `catch` statement and not tripping the outer `catch` statement.

Yeah, Julia's scoping rules can be pretty unintuitive, especially coming from Python. I think the modifications to `read_demand_flexibility` make it much clearer.

danielolsen (approved):
Thanks. There's one place where a function gets extra parameters that it doesn't use, once this is removed I think this is ready to merge.

src/model.jl
lanesmith: Done.

lanesmith:
> Thanks. There's one place where a function gets extra parameters that it doesn't use, once this is removed I think this is ready to merge.

Great, thank @BainanXia and @danielolsen for the feedback!

Issue #80, opened 2020-11-04
Add CI for Julia tests


danielolsen:
As mentioned in https://github.com/Breakthrough-Energy/REISE.jl/pull/79

Background: https://github.com/marketplace/actions/setup-julia-environment

The file that we will want to run: https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/test/runtests.jl. As we add more Julia side tests, we will keep including them in this file so that they all get triggered.

PR #120, opened 2021-04-05
Add julia test CI

danielolsen:
### Purpose
Addresses #80.

### What is the code doing
On push or PR, runs tests on Julia versions {1.5, latest}. File generated using `PkgTemplates.jl`, default settings for `PkgTemplates.GitHubActions`, with Julia version 1.0 removed manually from the version matrix (our current dependencies are apparently not compatible with 1.0, see https://github.com/Breakthrough-Energy/REISE.jl/runs/2267145710).

### Time to review
5 minutes? I don't understand github CI yaml very well but you probably do. Maybe we want to only test on push to match the python tests?


jon-hagg:
The trade off with `pull_request` trigger is we get duplicate runs within a PR, but it's required if we want actions to run on PR's from forks. I might leave it out for now unless we expect external contributions, but either way is fine.

.github/workflows/test-julia.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
```
jon-hagg, line 1: We should either have the names reflect julia/python, or combine the jobs into one workflow called "CI". I'm leaning slightly toward the latter, any thoughts?

danielolsen: Oops, I guess I forgot to push that change. I had changed it to `Test` locally, but never committed/pushed.

Combined testing jobs are fine with me. So we would just have **test.yml**, with name `____`, with two jobs names e.g. `test-python` and `test-julia`?

danielolsen:
> The trade off with `pull_request` trigger is we get duplicate runs within a PR, but it's required if we want actions to run on PR's from forks. I might leave it out for now unless we expect external contributions, but either way is fine.

I'm fine with leaving it on `push` only until we get our first external PR, if/when that happens we can update the settings and ask them to rebase to the latest.

.github/workflows/test-julia.yml
jon-hagg: Yeah, sounds good! I suppose naming it `Test` also makes more sense than `CI`.

.github/workflows/test-julia.yml
```diff
@@ -0,0 +1,36 @@
...
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
```
jon-hagg, line 27: Do we need both `cache-name` and the `-test-` component? I'm thinking these could be combined/shortened. Maybe something like `${{ runner.os }}-artifacts-${{ hashFiles('**/Project.toml') }}`.

.github/workflows/test-julia.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
```
jon-hagg, line 33: I think the last two keys are probably unnecessary

danielolsen: I trust you, I have no idea what these do. I will remove.

.github/workflows/test-julia.yml
danielolsen: I have no idea. I'll change it per your suggestion.

.github/workflows/test-julia.yml
danielolsen: Done.

.github/workflows/test-julia.yml
danielolsen: Done.

jon-hagg (approved):
This is great, thanks for setting it up.

PR #121, opened 2021-04-05
ci: use similar yaml across jobs, simplify naming

jon-hagg:
Minor improvements:
* remove `arch` from matrix since I don't expect we'll use other values (e.g. x86)
* leave out the os from the job name for now (we can add it back if we start running on more than just ubuntu)
* fix the `restore-keys` list - my understanding is that setting the `cache-name` environment variable is used as an example in the docs to show cases like having multiple caches based on feature branches or some other variable. we aren't partitioning it this way (and I'm not super clear on when that would be useful anyway), so having a hard coded `artifacts` is sufficient. 

Feel free to cherry pick or squash if it looks good.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,8 @@ on: push
 
 jobs:
   test-julia:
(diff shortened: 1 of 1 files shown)


danielolsen (approved):
I trust you

PR #122, opened 2021-04-05
test: reduce noise from flaky test

jon-hagg:
### Purpose
The object being tested involves spawning a process, along with a background thread, so inherently has a race condition. This has been historically semi stable but to prevent noise we have it sleep, to give the background thread some extra time. The duration is arbitrary, just didn't want to make it too long. 

### Testing
Ideally we would measure the failure rate over time, but this will have to be manually observed until there is a way to do that.

### Time to review
30 sec

### DIFF
--- a/pyreisejl/utility/tests/test_state.py
+++ b/pyreisejl/utility/tests/test_state.py
@@ -1,3 +1,4 @@
+import time
 from subprocess import PIPE, Popen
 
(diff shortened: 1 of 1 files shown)


danielolsen (approved):
Fingers crossed.

rouille (approved):
Makes sense

PR #130, opened 2021-05-06
refactor: make api available outside flask context

jon-hagg:
### Purpose
Enable these functions to be called outside of flask, so a user can launch a simulation and check the status without running the web server, which is mainly useful for communication between containers.

### What it does
Separate the functions from flask  (no route decorator or usage of `jsonify`, which creates an http response object).

By adding the path to REISE.jl to `sys.path`, we can import this module within powersimdata.

### Testing
Ongoing, along with the Breakthrough-Energy/PowerSimData#477

### Time to review
5 mins

### DIFF
--- a/pyreisejl/utility/app.py
+++ b/pyreisejl/utility/app.py
@@ -26,11 +26,8 @@ def get_script_path():
     return str(path_to_script)
 
 
-@app.route("/launch/<int:scenario_id>", methods=["POST"])
-def launch_simulation(scenario_id):
+def launch_simulation(scenario_id, threads=None, solver=None):
     cmd_call = ["python3", "-u", get_script_path(), str(scenario_id), "--extract-data"]
-    threads = request.args.get("threads", None)
-    solver = request.args.get("solver", None)
 
     if threads is not None:
         cmd_call.extend(["--threads", str(threads)])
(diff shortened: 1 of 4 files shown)


pyreisejl/utility/app.py
```diff
@@ -26,11 +26,8 @@ def get_script_path():
     return str(path_to_script)
 
 
-@app.route("/launch/<int:scenario_id>", methods=["POST"])
-def launch_simulation(scenario_id):
+def launch_simulation(scenario_id, threads=None, solver=None):
```
ahurli, line 29: If we're going to be using this to launch simulations from `powersimdata`, should we add a check for available solvers here just so that it fails faster? i.e. `from pyreisejl.utility.launchers import get_available_solvers`?

ahurli (approved):

ahurli:
Looks good! I'm assuming to the `poxispath` came about because you were testing it with the `powersimdata` changes on windows, and now it works for windows too, right?

jon-hagg:
> Looks good! I'm assuming to the `poxispath` came about because you were testing it with the `powersimdata` changes on windows, and now it works for windows too, right?

Yep, exactly!

pyreisejl/utility/const.py
```diff
@@ -1,9 +1,13 @@
...
-SCENARIO_LIST = posixpath.join(DATA_ROOT_DIR, "ScenarioList.csv")
-EXECUTE_LIST = posixpath.join(DATA_ROOT_DIR, "ExecuteList.csv")
-EXECUTE_DIR = posixpath.join(DATA_ROOT_DIR, "tmp")
-INPUT_DIR = posixpath.join(DATA_ROOT_DIR, "data", "input")
-OUTPUT_DIR = posixpath.join(DATA_ROOT_DIR, "data", "output")
+SCENARIO_LIST = os.path.join(DATA_ROOT_DIR, "ScenarioList.csv")
+EXECUTE_LIST = os.path.join(DATA_ROOT_DIR, "ExecuteList.csv")
+EXECUTE_DIR = os.path.join(DATA_ROOT_DIR, "tmp")
+INPUT_DIR = os.path.join(DATA_ROOT_DIR, "data", "input")
+OUTPUT_DIR = os.path.join(DATA_ROOT_DIR, "data", "output")
```
rouille, line 13: One day, it will be nice to find a ways to not set these variables when we use **PowerSimData** to launch **REISE.jl**. What is particularly ugly is that the path needs to be identical to the ones set in **PowerSimData**

rouille (approved):
Looks good

Issue #129, opened 2021-05-06
Error when launching a simulation in container


jon-hagg:
I get this stack trace when launching a simulation using the demo script in plug. 

```
'File case_storage.mat not found in /mnt/bes/pcm/tmp/scenario_1',
  'Demand flexibility profiles not found in /mnt/bes/pcm/tmp/scenario_1',
  'Julia exception: UndefKeywordError: keyword argument duration not assigned',
  'Stacktrace:',
  '[1] read_demand_flexibility(::String) at /app/src/read.jl:178',
  '[2] run_scenario(; num_segments::Int64, interval::Int64, n_interval::Int64, start_index::Int64, inputfolder::String, outputfolder::String, threads::Nothing, optimizer_factory::Type{T} where T, solver_kwargs::Nothing) at /app/src/REISE.jl:64',
```

danielolsen:
Oops. This probably came from https://github.com/Breakthrough-Energy/REISE.jl/pull/118.

lanesmith:
Whoops, I'm working to address this right now.

BainanXia:
It's here: https://github.com/Breakthrough-Energy/REISE.jl/blob/f192da66d87c61d7e2c1fd9c1d8af5a3886873e7/src/read.jl#L105 Another csv is needed as inputs based on the latest PR @danielolsen mentioned above. `PowerSimData` hasn't been updated to support the latest demand flexibility feature yet as discussed with @lanesmith before the above PR is merged.

lanesmith:
> It's here:
> 
> https://github.com/Breakthrough-Energy/REISE.jl/blob/f192da66d87c61d7e2c1fd9c1d8af5a3886873e7/src/read.jl#L105
> 
> Another csv is needed as inputs based on the latest PR @danielolsen mentioned above. `PowerSimData` hasn't been updated to support the latest demand flexibility feature yet as discussed with @lanesmith before the above PR is merged.

There is a problem on the `REISE.jl` side. If a file is not provided, not all of the parameters in the `DemandFlexibility` are assigned. I thought that not all parameters need to be assigned, but I was wrong. There may still be problems on the `PowerSimData` side, but the error that @jon-hagg posted above matches with the `REISE.jl` error I just found. I'm making the REISE.jl fix right now and should be able to get a PR submitted soon.

PR #131, opened 2021-05-07
fix: specify all parameters for DemandFlexibility struct

lanesmith:
### Purpose

The purpose of this code is to fix the error discussed in #129. When a `demand_flexibility.csv` file is not available, not all of the parameters of the `DemandFlexibility` struct were defined.

### What is the code doing?

In the `try/catch` loop that is responsible for trying to load the `demand_flexibility.csv` file in `read_demand_flexibility`, the three parameters that were previously unspecified are now specified.

### Where to look

The fix lives in the `read_demand_flexibility` function in `src/read.jl`.

### Testing

I tested the fix on my local machine and the error presented in #129 no longer occurs. As @BainanXia mentioned, the demand flexibility functionality is not yet integrated into `PowerSimData`, so I cannot be sure that there are not any errors on that end. I also have not tested this with `plug`, which is where the error was first noted. However, based on the error reported in #129, I believe this fix should do the trick.

### Time estimate

This review should be pretty quick.

### DIFF
--- a/src/read.jl
+++ b/src/read.jl
@@ -117,6 +117,9 @@ function read_demand_flexibility(filepath)::DemandFlexibility
         println("Demand flexibility profiles not found in " * filepath)
         demand_flexibility["flex_amt"] = nothing
         demand_flexibility["enabled"] = false
+        demand_flexibility["duration"] = nothing
+        demand_flexibility["interval_balance"] = false
+        demand_flexibility["rolling_balance"] = false
     end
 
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
Looks good to me. Maybe @jon-hagg could give a quick try.

jon-hagg (approved):
Works for me, thanks!

lanesmith:
Great, I'll merge it in!

PR #135, opened 2021-06-20
feat: add flexible demand cost

### LINKED ISSUE #124: Add cost components to demand flexibility model
Currently, demand flexibility is modeled as being purely price responsive. Coupled with load balance constraints and bounds on the amount that demand can deviate (via demand flexibility profiles), this representation figures to model the impact of market-based demand response programs, utility time-of-use pricing programs, and even more traditional demand response programs (to an extent) fairly well. That being said, there could be a benefit in allowing additional cost parameters to be specified. Two such costs that have been mentioned initially include:

- Curtailment cost associated with reducing demand. This could potentially be thought of as an inconvenience cost.
- A bid curve, similar to what is seen with the generators and closer to what might be expected in a market setting.

A better determination of the cost parameters that should be included will likely be able to be made once BE's sectoral demand collaborators get to the demand flexibility portions of their projects. Since NREL does not include cost parameters with the [demand flexibility profiles](https://data.nrel.gov/submissions/127) used in their [Electrification Futures Study](https://www.nrel.gov/analysis/electrification-futures.html), we do not have cost parameters to pair with our initial demand flexibility profiles.

dongqi-wu:
- Modified the DemandFlexibility type in types.jl to include two containers for cost table
- Modified read.jl to read two additional csvs from the input table named "demand_flexibility_cost_up.csv" and "demand_flexibility_cost_dn.csv"
- Modified model.jl to impost a cost on decision variables "load_shift_up" and "load_shift_dn"

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -170,7 +170,7 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
     # Storage
-    storage_enabled = isa(storage, Storage) && (size(storage.gen, 1) > 0)
+    storage_enabled = isa(storage, Storage) && storage.enabled
     num_storage = storage_enabled ? size(storage.gen, 1) : 0
     storage_idx = storage_enabled ? (1:num_storage) : nothing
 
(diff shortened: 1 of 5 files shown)


danielolsen:
The code looks straightforward. Do you have sample data that you've used for testing?

dongqi-wu:
> The code looks straightforward. Do you have sample data that you've used for testing?

Hi Daniel,

Yes, I have the files that can be directly used for testing. Shall I upload them to somewhere? I don't see a corresponding folder in REISE.jl so I didn't push the files.

danielolsen:
> > The code looks straightforward. Do you have sample data that you've used for testing?
> 
> Hi Daniel,
> 
> Yes, I have the files that can be directly used for testing. Shall I upload them to somewhere? I don't see a corresponding folder in REISE.jl so I didn't push the files.

Yes, if you can upload the required files somewhere that would be helpful. If you are testing with one of the base grids from PowerSimData, no need to upload the Grid or the base demand/hydro/solar profiles, since we can get them from the repo.

src/model.jl
```diff
@@ -101,6 +101,30 @@ function _make_bus_demand_flexibility_amount(
...
+
+Given a matrix of demand response cost table, select a sub-matrix of demand flexibility cost
+    by (hour, num_load_bus) for this interval.
+"""
+function _make_bus_demand_flexibility_cost_amount(
+    demand_flexibility_cost::Union{DataFrames.DataFrame,Nothing}, start_index::Int, end_index::Int, num_load_bus::Int
+)::Matrix
+    # check if the cost table exist, if not, create a DataFrame of zeros
+    if isnothing(demand_flexibility_cost)
+        bus_demand_flex_cost_amt = zeros(Int8, end_index - start_index + 1, num_load_bus)
```
lanesmith: We might not need to create a `DataFrame` of zeros. Instead, we could use `if` statements to check all the places that would need the demand flexibility cost (e.g., later in `model.jl` where `_make_bus_demand_flexibility_cost_amount` is actually called and where the objective function is amended to include the demand flexibility cost) to see if `demand_flexibility.cost_up` and `demand_flexibility.cost_dn` are `nothing`. If they are not `nothing` (and I imagine they should be checked individually), then we could call this function or implement the objective function amendment.

src/model.jl
```diff
@@ -261,6 +285,12 @@ function _build_model(
         bus_demand_flex_amt = _make_bus_demand_flexibility_amount(
             case, demand_flexibility, start_index, end_index
         )
+        bus_demand_flex_cost_up_amt = _make_bus_demand_flexibility_cost_amount(
+            demand_flexibility.cost_up, start_index, end_index, sets.num_load_bus
+        )
```
lanesmith: If we follow my suggestion from above, then this should be preceded by something like `if !isnothing(demand_flexibility.cost_up)` and would need to be followed with `end`.

src/model.jl
```diff
@@ -261,6 +285,12 @@ function _build_model(
...
+        bus_demand_flex_cost_dn_amt = _make_bus_demand_flexibility_cost_amount(
+            demand_flexibility.cost_dn, start_index, end_index, sets.num_load_bus
+        )
```
lanesmith: Similar to above comment.

src/model.jl
```diff
@@ -528,6 +558,23 @@ function _build_model(
...
         JuMP.add_to_expression!(obj, storage_penalty)
     end
+    
+    # Pay for the cost of DR programs based on committed flexible demand
+    if demand_flexibility.enabled
+        
+        # cost for increasing flexible load
+        demand_response_penalty_up = JuMP.@expression(m, sum(
+            sum(bus_demand_flex_cost_up_amt * load_shift_up)))
+        JuMP.add_to_expression!(obj, demand_response_penalty_up)
```
lanesmith: If we follow my suggestion from above, then this should be preceded by something like if `!isnothing(demand_flexibility.cost_up)` and would need to be followed with `end`.

src/model.jl
```diff
@@ -528,6 +558,23 @@ function _build_model(
...
+        
+        # cost for decreasing flexible load
+        demand_response_penalty_dn = JuMP.@expression(m, sum(
+            sum(bus_demand_flex_cost_dn_amt * load_shift_dn)))
+        JuMP.add_to_expression!(obj, demand_response_penalty_dn)
```
lanesmith: Similar to above comment.

src/read.jl
```diff
@@ -122,12 +122,42 @@ function read_demand_flexibility(filepath)::DemandFlexibility
...
+    
+    
     # Set the demand flexibility parameters
     if demand_flexibility["enabled"]
         # Pre-specify the demand flexibility parameters
         demand_flexibility["duration"] = nothing
         demand_flexibility["interval_balance"] = true
         demand_flexibility["rolling_balance"] = true
+    
+
```
lanesmith: I think some of these extra spaces through the commits in this PR can be cleaned up a little.

src/read.jl
```diff
@@ -122,12 +122,42 @@ function read_demand_flexibility(filepath)::DemandFlexibility
         demand_flexibility["rolling_balance"] = false
     end
 
+    
+    # Try loading demand flexibility cost. If a file is not found, set the field to nothing.
+    if demand_flexibility["enabled"]
```
lanesmith: This looks good, though for brevity the code in this logic could probably be added to the previous block of logic, where `demand_flexibility["enabled"]` is already checked. If you make this change, be sure that this lies outside the `try`-`catch` check.

lanesmith (commented):
Overall, I think it's looking good! Please see my comments, which mostly address checking whether `demand_flexibility.cost_up` and `demand_flexibility_cost_dn` are `nothing` rather than creating a `DataFrame` of zeros.

lanesmith:
I was just thinking that we might also need to update `loop.jl` so that the objective function is properly updated on intervals `i > 2`. If `demand_flexibility.cost_up` and `demand_flexibility.cost_dn` were time invariant, then I don't think this would be a problem. I know the test data is time invariant right now, but it it set up to be time varying in the future, right? If so, I think we'll need to update the `demand_response_penalty_up` and `demand_response_penalty_dn` expressions for each interval (so as to pass in the proper `bus_demand_flex_cost_up_amt` and `bus_demand_flex_cost_dn_amt` values, respectively).

BainanXia:
> I was just thinking that we might also need to update `loop.jl` so that the objective function is properly updated on intervals `i > 2`. If `demand_flexibility.cost_up` and `demand_flexibility.cost_dn` were time invariant, then I don't think this would be a problem. I know the test data is time invariant right now, but it it set up to be time varying in the future, right? If so, I think we'll need to update the `demand_response_penalty_up` and `demand_response_penalty_dn` expressions for each interval (so as to pass in the proper `bus_demand_flex_cost_up_amt` and `bus_demand_flex_cost_dn_amt` values, respectively).

Good point. If that is the case, I presume we will want the cost properly defined for all time intervals starting from i==1, right? The reason we are looking for i > 2 right now is because it relates to ramp constraints only, which can't be defined for i==1. Maybe @danielolsen has a better idea on this.

danielolsen:
I agree. If I recall correctly, this is the first place in which objective functions coefficients change from interval to interval, rather than adding constraints (e.g. adding ramping constraints) or modifying the values of constraint parameters (e.g. load shedding, or updating initial conditions).

lanesmith:
> I agree. If I recall correctly, this is the first place in which objective functions coefficients change from interval to interval, rather than adding constraints (e.g. adding ramping constraints) or modifying the values of constraint parameters (e.g. load shedding, or updating initial conditions).

Yes, I believe so

lanesmith:
> If that is the case, I presume we will want the cost properly defined for all time intervals starting from i==1, right? 

The costs for `i = 1` and `i = 2` should be fine because a new JuMP model is created in each of those instances. The creation of a new model appears to properly define the costs in `model.jl`. We don't create a new JuMP model for `i > 2`, so the values will need to be respecified in those cases so that the correct temporal data is used.

BainanXia:
> > If that is the case, I presume we will want the cost properly defined for all time intervals starting from i==1, right?
> 
> The costs for `i = 1` and `i = 2` should be fine because a new JuMP model is created in each of those instances. The creation of a new model appears to properly define the costs in `model.jl`. We don't create a new JuMP model for `i > 2`, so the values will need to be respecified in those cases so that the correct temporal data is used.

Agree.

PR #137, opened 2021-06-25
ci: add PR trigger

jenhagg:
### Purpose
Allows workflows from forks to run automatically.

### What the code is doing
Add the trigger to workflows

### Testing
n/a (we already have this in PreREISE)

### Time estimate
2 min


rouille (approved):
Makes sense

danielolsen (approved):
Thanks.

Issue #60, opened 2020-07-08
Add selection for DCOPF, network flow, or copper plate transmission model


danielolsen:

danielolsen:
For a network/transport model, we would still have power flow on each line, it just wouldn't be constrained by KVL. But with a copper-plate model, we no longer have meaningful values for `pf`. In order to keep these scenarios compatible with PowerSimData, I think we either need to same an empty **pf.pkl** file in the extraction process and/or modify PowerSimData so that it knows how to properly interpret a missing `pf` file.

danielolsen:
Another transmission model we may be interested in: copper-plate, within interconnections only. I think we could enable this by removing all flow limits on AC lines, but not DC lines. This could let us investigate e.g. the difference between the best-case transmission upgrades within interconnections vs. the best case transmission upgrades including bridging the interconnections.

BainanXia:
> Another transmission model we may be interested in: copper-plate, within interconnections only. I think we could enable this by removing all flow limits on AC lines, but not DC lines. This could let us investigate e.g. the difference between the best-case transmission upgrades within interconnections vs. the best case transmission upgrades including bridging the interconnections.

This can be achieved by scaling the capacities of all AC branches to 0 via change table right, although, it is ugly in some sense.

danielolsen:
@BainanXia scaling the capacities of the AC branches down to zero will remove the min & max constraints on the AC lines, but REISE.jl will still add variables for the flow on each branch, enforce branch-angle constraints on their flows, and create power-balance constraints for each bus. It would work, but still creates a model with unnecessary complexity (although maybe Gurobi would realize this and presolve some/all of the unnecessary complexity away).

I think what we want is:
- DCOPF: branch-angle, flow min/max, power balance
- network flow: flow min/max, power balance
- interconnection copper-plate: flow min/max on DC only, power balance (only really necessary at the seams, but I think we might need more complex code to automatically collapse AC interconnections to single nodes)
- copper-plate: generation = demand

BainanXia:
@danielolsen Agree. A true copper-plate mode should ignore the topology completely as long as the graph is a connected component.

danielolsen:
[...]

PR #142, opened 2021-07-09
refactor: simplify _build_model by defining functions that add constraints

### LINKED ISSUE #117: Refactor model building function for simplicity
Currently, the `_build_model` function is ~260 lines, about half of which is the building of constraints. We should be able to reduce the complexity of this function by creating smaller functions to add each constraint, or each family of constraints, to keep `_build_model` to a more reasonable size.

An attempt was made in an old branch (see the last 6 commits of https://github.com/Breakthrough-Energy/REISE.jl/commits/constraint_functions) but it's pretty out of date, so we should probably start fresh and make better use of our Sets struct.

danielolsen:

### Purpose
The `_build_model` function is currently 360 lines long, and will only get more complicated as we add more functionality. This PR simplifies this function by moving logic to add constraints to a model to individual functions. The end result is that the length of `_build_model` is reduced to 200 lines. This can serve other refactoring efforts, e.g. #60, as well as improve the test coverage, since now it is easier to write unit tests to ensure that each constraint or set of constraints is being build properly.
Closes #117.

Separately, we also implement a fix to the demand flexibility implementation, which I found during testing.

### What the code is doing
We add new functions:
- `_add_constraint_power_balance!`
- `_add_constraint_load_shed!`
- `_add_constraints_storage_operation!`
- `_add_constraints_demand_flexibility!`
- `_add_constraints_initial_ramping!`
- `_add_constraints_ramping!`
- `_add_constraints_generator_segments!`
- `_add_constraints_branch_flow_limits!`
- `_add_branch_angle_constraints!`
- `_add_profile_generator_limits!`
- `_add_objective_function!`

Per Julia convention, these function names end with `!` since they modify at least one of the inputs (the JuMP model).

Within these functions, we make use of the JuMP syntax model[:foo] to access the variable or contraint (or array of variables/contraints) named foo that was defined in the model object. This way, we can simplify passing the model object back and forth, and don't have to manually plumb the foo references that get created in the _build_model namespace.

### Testing
Tested manually with storage, load shedding, and demand flexibility. Given the scope of this refactor, addition testing by other people is highly appreciated.

### Usage Example/Visuals
Usage from user-facing functions is unchanged.

### Time estimate
\>= 1 hour. There is still work to be done in improving the format of the code, but the main thing I'm looking for from reviewers at this draft stage is an assessment of the organization.

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -565,7 +565,7 @@ function _build_model(
     initial_ramp_enabled::Bool=false,
     initial_ramp_g0::Array{Float64,1}=Float64[],
     storage_e0::Array{Float64,1}=Float64[]
-)::Tuple{JuMP.Model, VariablesOfInterest}
+)::JuMP.Model
     # Positional indices from mpc.gen
     PMAX = 9
     PMIN = 10
@@ -696,26 +696,5 @@ function _build_model(
     )
 
     println(Dates.now())
-    # For non-existent variables/constraints, define as `nothing`
-    load_shed = load_shed_enabled ? load_shed : nothing
-    load_shift_up = demand_flexibility.enabled ? load_shift_up : nothing
-    load_shift_dn = demand_flexibility.enabled ? load_shift_dn : nothing
-    storage_dis = storage_enabled ? storage_dis : nothing
-    storage_chg = storage_enabled ? storage_chg : nothing
-    storage_soc = storage_enabled ? storage_soc : nothing
-    initial_soc = storage_enabled ? initial_soc : nothing
-    initial_rampup = initial_ramp_enabled ? initial_rampup : nothing
-    initial_rampdown = initial_ramp_enabled ? initial_rampdown : nothing
-    load_shed_ub = load_shed_enabled ? load_shed_ub : nothing
-    voi = VariablesOfInterest(;
-        # Variables
-        pg=pg, pf=pf, 
-        load_shed=load_shed, load_shift_up=load_shift_up, load_shift_dn=load_shift_dn, 
-        storage_soc=storage_soc, storage_dis=storage_dis, storage_chg=storage_chg,
-        # Constraints
-        branch_min=branch_min, branch_max=branch_max, powerbalance=powerbalance,
-        initial_soc=initial_soc, load_shed_ub=load_shed_ub, 
-        initial_rampup=initial_rampup, initial_rampdown=initial_rampdown,
-        hydro_fixed=hydro_fixed, solar_max=solar_max, wind_max=wind_max)
-    return (m, voi)
+    return m
 end
--- a/src/model.jl
+++ b/src/model.jl
@@ -170,7 +170,7 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
     # Storage
-    storage_enabled = isa(storage, Storage) && (size(storage.gen, 1) > 0)
+    storage_enabled = isa(storage, Storage) && storage.enabled
     num_storage = storage_enabled ? size(storage.gen, 1) : 0
     storage_idx = storage_enabled ? (1:num_storage) : nothing
 
@@ -211,8 +211,7 @@ function _add_constraint_power_balance!(
         injections = JuMP.@expression(m, injections + sets.load_bus_map * m[:load_shed])
     end
     withdrawals = JuMP.@expression(m, bus_demand)
-    storage_enabled = (sets.num_storage > 0)
-    if storage_enabled
+    if storage.enabled
         storage_bus_idx = [sets.bus_id2idx[b] for b in storage.gen[:, 1]]
         storage_map = sparse(
             storage_bus_idx, sets.storage_idx, 1, sets.num_bus, sets.num_storage
@@ -497,7 +496,6 @@ function _add_objective_function!(
     load_shed_enabled::Bool,
     load_shed_penalty::Number,
     trans_viol_enabled::Bool,
-    storage_enabled::Bool,
     storage_e0::Array{Float64,1},
 )
     # Positional indices from mpc.gencost
@@ -528,7 +526,7 @@ function _add_objective_function!(
         )
     end
     # Pay for ending with less storage energy than initial
-    if storage_enabled
+    if storage.enabled
         storage_penalty = JuMP.@expression(
             m,
             sum(
@@ -579,7 +577,6 @@ function _build_model(
 
     println("parameters: ", Dates.now())
     # Parameters
-    storage_enabled = (sets.num_storage > 0)
     bus_demand = _make_bus_demand(case, start_index, end_index) * demand_scaling
     # Demand flexibility parameters (if present)
     if demand_flexibility.enabled
@@ -609,7 +606,7 @@ function _build_model(
             0 <= trans_viol[i in sets.branch_idx, j in hour_idx],
             container=Array)
     end
-    if storage_enabled
+    if storage.enabled
         storage_max_dis = storage.gen[:, PMAX]
         storage_max_chg = -1 * storage.gen[:, PMIN]
         storage_min_energy = storage.sd_table.MinStorageLevel
@@ -652,7 +649,7 @@ function _build_model(
         _add_constraint_load_shed!(m, case, sets, demand_flexibility, bus_demand)
     end
 
-    if storage_enabled
+    if storage.enabled
         _add_constraints_storage_operation!(
             m, case, sets, storage, interval_length, storage_e0
         )
@@ -691,7 +688,6 @@ function _build_model(
         load_shed_enabled,
         load_shed_penalty,
         trans_viol_enabled,
-        storage_enabled,
         storage_e0,
     )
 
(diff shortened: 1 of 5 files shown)


danielolsen:
@dongqi-wu I think your feedback here would also be useful, since you'll be working with this codebase. I am agnostic as to the merge order between this PR and #135; if yours gets merged first, I can rebase around your changes without too much work.

BainanXia:
This makes perfect sense to me and the logic is clean. Will read through the code once the PR is formally setup. Thanks.

danielolsen:
I've made some changes to the code style to make it a bit easier to read. I haven't done too much research, but it seems like a Julia analog to python's `black` code style is Blue: https://github.com/invenia/BlueStyle, and I've made some change that I think conform to that standard. We will probably want a more thorough style cleanup later for the rest of the codebase.

I think this is now ready for a full review.

dongqi-wu:
> @dongqi-wu I think your feedback here would also be useful, since you'll be working with this codebase. I am agnostic as to the merge order between this PR and #135; if yours gets merged first, I can rebase around your changes without too much work.

Hi Daniel,

I have looked at the changes, I think is a very good idea to separate build_model into individual functions so it becomes easier to change and test them. For my PR, I have addressed and tested Lane's comments locally and I will override my commit on PR soon. We can discuss it Tuesday, if there is no more issues for the new version we can probably merge it shortly after the meeting.

BainanXia:
> > @dongqi-wu I think your feedback here would also be useful, since you'll be working with this codebase. I am agnostic as to the merge order between this PR and #135; if yours gets merged first, I can rebase around your changes without too much work.
> 
> Hi Daniel,
> 
> I have looked at the changes, I think is a very good idea to separate build_model into individual functions so it becomes easier to change and test them. For my PR, I have addressed and tested Lane's comments locally and I will override my commit on PR soon. We can discuss it Tuesday, if there is no more issues for the new version we can probably merge it shortly after the meeting.

Hi Dongqi,

Thanks for the update. Please hold off overriding you commit on the open PR. We haven't heard from you last week and thought you didn't have access to Git in China. @lanesmith has prepared a draft PR based on your branch to address his comments and we are going to talk about it during the working session right after our next meeting (we will go through your local changes as well). Ideally, we can get it merged during the working session.

dongqi-wu:
> > > @dongqi-wu I think your feedback here would also be useful, since you'll be working with this codebase. I am agnostic as to the merge order between this PR and #135; if yours gets merged first, I can rebase around your changes without too much work.
> > 
> > 
> > Hi Daniel,
> > I have looked at the changes, I think is a very good idea to separate build_model into individual functions so it becomes easier to change and test them. For my PR, I have addressed and tested Lane's comments locally and I will override my commit on PR soon. We can discuss it Tuesday, if there is no more issues for the new version we can probably merge it shortly after the meeting.
> 
> Hi Dongqi,
> 
> Thanks for the update. Please hold off overriding you commit on the open PR. We haven't heard from you last week and thought you didn't have access to Git in China. @lanesmith has prepared a draft PR based on your branch to address his comments and we are going to talk about it during the working session right after our next meeting (we will go through your local changes as well). Ideally, we can get it merged during the working session.

Hi Bainan,

Sure, I'm sorry that I didn't let you know in advance that I was already working on the comments. I already have a VPN through which I can access anything including git. We can compare our change on Tuesday and hopefully merge the better version.

src/model.jl
```diff
@@ -190,6 +190,369 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
...
+    bus_demand *= demand_scaling
+
+    gen_injections = JuMP.@expression(m, gen_map * m[:pg])
+    line_injections = JuMP.@expression(m, branch_map * m[:pf])
+    injections = JuMP.@expression(m, gen_injections + line_injections)
+    if load_shed_enabled
+        injections = JuMP.@expression(m, injections + sets.load_bus_map * m[:load_shed])
+    end
+    withdrawals = JuMP.@expression(m, bus_demand)
+    storage_enabled = (sets.num_storage > 0)
```
lanesmith: Can `storage_enabled` be passed in as an input? It looks like it is already created and used elsewhere in `_build_model`.

src/model.jl
```diff
@@ -190,6 +190,369 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
...
+    demand_scaling::Number,
+    load_shed_enabled::Bool,
+)
+    # Generator topology matrix
+    gen_map = _make_gen_map(case)
+    # Branch connectivity matrix
+    branch_map = _make_branch_map(case)
+    # Demand by bus
+    bus_demand = _make_bus_demand(case, start_index, start_index + interval_length - 1)
+    bus_demand *= demand_scaling
```
lanesmith: Can `bus_demand` be passed in as an input (following the demand scaling)? It looks like it is also needed for `_add_constraint_load_shed!` (if load shedding is enabled). When load shedding is not enabled, `bus_demand` is only created once, as it currently does. When load shedding is enabled, we would reduce one of the times that `bus_demand` is created. If we do this change, it appears that we would no longer need `start_index` and `demand_scaling` as inputs.

src/model.jl
```diff
@@ -190,6 +190,369 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
...
+    case::Case,
+    sets::Sets,
+    demand_flexibility::DemandFlexibility,
+    start_index::Int,
+    interval_length::Int,
+    demand_scaling::Number,
+)
+    # Demand by bus
+    end_index = start_index + interval_length - 1
+    bus_demand = _make_bus_demand(case, start_index, end_index) * demand_scaling
```
lanesmith: Similar to above comment for `add_constraint_power_balance!`.

src/model.jl
```diff
@@ -527,19 +728,20 @@ function _build_model(
...
     storage_soc = storage_enabled ? storage_soc : nothing
-    initial_soc = storage_enabled ? initial_soc : nothing
-    initial_rampup = initial_ramp_enabled ? initial_rampup : nothing
-    initial_rampdown = initial_ramp_enabled ? initial_rampdown : nothing
-    load_shed_ub = load_shed_enabled ? load_shed_ub : nothing
+    initial_soc = storage_enabled ? m[:initial_soc] : nothing
+    initial_rampup = initial_ramp_enabled ? m[:initial_rampup] : nothing
+    initial_rampdown = initial_ramp_enabled ? m[:initial_rampdown] : nothing
+    load_shed_ub = load_shed_enabled ? m[:load_shed_ub] : nothing
     voi = VariablesOfInterest(;
```
lanesmith: I know we use the `VariablesOfInterest` struct in both `loop.jl` and `query.jl`, but I'm curious if using the syntax that was mentioned at the beginning of the PR might work too (i.e., calling `m[:foo]` for variable/constraint `foo` whenever needed). Since the model is already present, we would be accessing something that is already stored in memory. Do you know if this would work or if there would be any benefit in doing this over the current method?

lanesmith (commented):
Overall, I think it looks great! `_build_model` is definitely much cleaner. After addressing my comments, I don't see a reason that this shouldn't be approved, though I'll give the other reviewers more time.

src/model.jl
danielolsen: I thought about that. I guess it's a question of whether we want minimal inputs (i.e. infer whether storage is enabled from `sets`) or minimal code (avoid this line because we already know from the parameter being explicitly passed). I'm agnostic on this, and would follow your and @BainanXia's advice here.

I guess this applies to your next two comments as well.

src/model.jl
danielolsen: That might work! The more I look at `VariablesOfInterest`, the less I like it. It would definitely be cleaner code to just look directly into the model itself, now that we know the convenient syntax to do so.

src/model.jl
lanesmith: I hadn't considered the minimal input vs minimal code tradeoff. Either way is fine to me. What I was thinking when I left the comment was that passing in `storage_enabled` would reduce the number of computations (though only marginally so, I imagine). If we really wanted to reduce the number of times that `storage_enabled` is calculated, we could probably add an `enabled` member to the `Storage` struct (similar to what we do with `DemandFlexibility`).

While this is similar to the next two comments, I believe we would have input savings by passing in a once-calculated `bus_demand`.

BainanXia: I think for a simple and compact model, we can implement in a way that it could infer what to do based on the inputs, however, as we keep adding new features to our model, it would be better to let user be explicit on what he/she would like to do via more input parameters. Hence, I lean towards the latter at this point. Any inputs from @jon-hagg @rouille ?

danielolsen: Long-term, I think `Storage.enabled`, mimicing what we do in `DemandFlexibility`, is definitely the way to go.

src/model.jl
danielolsen: This is implemented in the latest batch of commits, and _seems_ to work...

lanesmith: Looks great!

src/model.jl
danielolsen: `Storage.enabled` has been implemented.

src/model.jl
danielolsen: Good call, done.

src/model.jl
danielolsen: Done.

BainanXia (approved):
Still getting familiar with Julia syntax though. Reading through the code, the logic makes sense to me. Any final comments @lanesmith ?

lanesmith (approved):
Looks good to me too! Thank you for addressing my comments

Issue #125, opened 2021-04-14
Add a representation of load shedding to the demand flexibility model


lanesmith:
Currently, all demand flexibility is implemented as "shiftable demand", where all curtailed demand must be recovered. However, flexible demand can also be classified as "sheddable demand", where curtailed demand does not have to be recovered. Sheddable demand is common in incentive-based demand response programs. Such a representation would be useful, especially considering the demand response with human-in-the-loop collaboration.

Implementing such a model would likely be easy, especially with much of the code infrastructure in place from the current shiftable demand model. Sheddable demand would be very similar, with the exceptions that (1) it would not need load balancing constraints and (2) it would not be able to be price responsive like the shiftable demand (due to the cost minimization, the sheddable demand would want to curtail its maximum amount every time step). For sheddable demand to adequately participate in the model, there will need to be a different representation of costs and benefits, as discussed in #124.

NOTE: Although load shedding does already exist in the model, this load shedding would be different (though not too different mathematically). While the current representation of load shedding is used mostly for feasibility purposes (e.g., extreme line congestion preventing a simulation from being feasible), the proposed load shedding discussed in this issue would be used to model the impact that demand response programs can have on the system.

merrielle:
@lanesmith Do we still need to work on this?

lanesmith:
> @lanesmith Do we still need to work on this?

Yeah, it's still not implemented. In the current way we model demand flexibility, which is solely through load shifting, load shedding would not work as intended because it would simply shed all demand that it could. Under the current framework, we will need to have costs associated with the load shedding so that shed decisions are more reasonable. We have a basic cost parameter associated with the demand flexibility model, but this is something that might change as our demand-response-with-human-in-the-loop collaborators continue to explore costs associated with demand flexibility. Unless we need this for something else, I would probably recommend holding off on this feature until the cost structure is better determined, just so we don't need to refactor too much in the future. However, if needed, it probably wouldn't be too difficult to get a basic model of load shedding implemented with the current cost framework.

PR #145, opened 2021-07-15
Separate demand flexibility profiles for up and down shifting

### LINKED ISSUE #139: Separate Demand Flexibility Data Input
# :rocket:

- [x] Is your feature request essential for your project?

I wish to split the input file for demand flexibility into two files that assign different allowance for shifting the demand upward or downward. This will enable a more flexible behaviour of flexible loads by allowing them to have different extend of flexibility in increasing or curtailing.

The script read.jl should take two csv files: 'demand_flexibility_up.csv' and 'demand_flexibility_dn.csv' instead of a single file 'demand_flexibility.csv'
The class "demand_flexibility" should have two dataframes to hold the data in the input csvs
The upper bound of the decision variables in the opt model 'load_shift_up' and 'load_shift_dn' should be different according to the two input files.

lanesmith:

### Purpose
The purpose of this PR is to create functionality to read and utilize two demand flexibility profiles: one that bounds the `load_shift_up` variable and another that bounds the `load_shift_dn` variable. This functionality will replace the existing functionality, where one demand flexibility profile is used as the bound for both `load_shift_up` and `load_shift_dn`. This PR closes #139 .

### What the code is doing
The majority of the changes are to the `read_demand_flexibility` function in `read.jl`. Here, the old functionality for reading a single demand flexibility profile has been converted into reading two demand flexibility profiles: `demand_flexibility_up.csv` and `demand_flexibility_dn.csv`. I also include a check that sets the `enabled` parameter to `true` if both demand flexibility profiles are present and `false` otherwise; `demand_flexibility_up.csv` not being present prevents demand recovery (since this is shiftable demand) and `demand_flexibility_dn.csv` not being present prevents curtailment. As mentioned in #139, if we want the ability to have sheddable demand, we should probably revisit #125. Also included in `read_demand_flexibility` is a small refactor that sets the `rolling_balance` parameter to `false` if the `duration` parameter equals the interval length; this refactor helps clean up some of the checks in `model.jl`.

The `DemandFlexibility` struct in `types.jl` is updated to replace the `flex_amt` parameters with `flex_amt_up` and `flex_amt_dn`.

In `model.jl`, the `_make_bus_demand_flexibility_amount` function was updated to be able to pass out two bus-level demand flexibility arrays. Both `model.jl` and `loop.jl` (1) reflect this change to `_make_bus_demand_flexibility_amount` and (2) feature updates to the upper bounds of the `load_shift_up` and `load_shift_dn` variables.

### Testing
Testing was conducted locally on the provided Texas data. The Texas data needed to be changed so that the old `demand_flexibility.csv` file is replaced by `demand_flexibility_up.csv` and `demand_flexibility_dn.csv`. For testing purposes, both `demand_flexibility_up.csv` and `demand_flexibility_dn.csv` are the same as `demand_flexibility.csv`. Under this definition of testing profiles, the results under the new implementation were the same as those under the old implementation.

### Where to look
The changes are contained in the following:
- The `interval_loop` function in `loop.jl`
- The `_make_bus_demand_flexibility_amount`, `_add_constraints_demand_flexibility!`, and `_build_model` functions in `model.jl`
- The `read_demand_flexibility` function in `read.jl`
- The `DemandFlexibility` struct in `types.jl`

### Time estimate
This hopefully shouldn't take too long to review as most of the changes are straightforward. Maybe 15 minutes?

### DIFF
--- a/src/read.jl
+++ b/src/read.jl
@@ -113,12 +113,12 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     for s in ["up", "dn"]
         try
             demand_flexibility["flex_amt_" * s] = CSV.File(
-                joinpath(filepath, "demand_flexibility_" * s * ".csv"))
+                joinpath(filepath, "demand_flexibility_" * s * ".csv")
             ) |> DataFrames.DataFrame
             println("...loading demand flexibility " * s * " profiles")
         catch e
-            println("Demand flexibility " * s " profile not found in " * filepath)
-            demand_flexibility[string("flex_amt_", s)] = nothing
+            println("Demand flexibility " * s * " profile not found in " * filepath)
+            demand_flexibility["flex_amt_" * s] = nothing
         end
     end
     
@@ -203,7 +203,14 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
             )
             demand_flexibility["duration"] = interval
         end
-        
+
+        # Prevent the rolling_balance constraint according to the duration parameter
+        if demand_flexibility["rolling_balance"]
+            if demand_flexibility["duration"] == interval
+                demand_flexibility["rolling_balance"] = false
+            end
+        end
+
         # Try reading the cost for up- and down-shifting loads
         for cost_file_suffix in ("up", "dn")
             try 
--- a/src/read.jl
+++ b/src/read.jl
@@ -128,6 +128,14 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     )
         demand_flexibility["enabled"] = true
     else
+        if !isnothing(demand_flexibility["flex_amt_up"]) || (
+            !isnothing(demand_flexibility["flex_amt_dn"])
+        )
+            @warn (
+                "The exclusion of one of the demand flexibility profiles has resulted "
+                * "in demand flexibility not being enabled."
+            )
+        end
         demand_flexibility["enabled"] = false
         demand_flexibility["duration"] = nothing
         demand_flexibility["interval_balance"] = false
--- a/src/read.jl
+++ b/src/read.jl
@@ -104,135 +104,156 @@ function read_storage(filepath)::Storage
 end
 
 
-"""Load demand flexibility profile from .csv files into DataFrame(s)."""
+"""
+    read_demand_flexibility(filepath, interval)
+
+Load demand flexibility profiles and parameters from .csv files and return them in a 
+DemandFlexibility struct.
+"""
 function read_demand_flexibility(filepath, interval)::DemandFlexibility
     # Initialize demand flexibility
-    demand_flexibility = Dict()
-
-    # Try loading demand flexibility profiles
-    for s in ["up", "dn"]
-        try
-            demand_flexibility["flex_amt_" * s] = CSV.File(
-                joinpath(filepath, "demand_flexibility_" * s * ".csv")
-            ) |> DataFrames.DataFrame
-            println("...loading demand flexibility " * s * " profiles")
-        catch e
-            println("Demand flexibility " * s * " profile not found in " * filepath)
-            demand_flexibility["flex_amt_" * s] = nothing
-        end
-    end
-    
-    # If both demand flexibility profiles exist, demand flexibility is enabled
-    if !isnothing(demand_flexibility["flex_amt_up"]) && (
-        !isnothing(demand_flexibility["flex_amt_dn"])
+    demand_flexibility = Dict(
+        "duration" => interval,
+        "enabled" => "not_specified",  
+        "interval_balance" => true, 
+        "rolling_balance" => true,
     )
-        demand_flexibility["enabled"] = true
-    else
-        if !isnothing(demand_flexibility["flex_amt_up"]) || (
-            !isnothing(demand_flexibility["flex_amt_dn"])
-        )
-            @warn (
-                "The exclusion of one of the demand flexibility profiles has resulted "
-                * "in demand flexibility not being enabled."
-            )
-        end
-        demand_flexibility["enabled"] = false
-        demand_flexibility["duration"] = nothing
-        demand_flexibility["interval_balance"] = false
-        demand_flexibility["rolling_balance"] = false
-        demand_flexibility["cost_up"] = nothing
-        demand_flexibility["cost_dn"] = nothing
-    end
 
-    # Set the demand flexibility costs and parameters
-    if demand_flexibility["enabled"]
-        # Pre-specify the demand flexibility parameters
-        demand_flexibility["duration"] = nothing
-        demand_flexibility["interval_balance"] = true
-        demand_flexibility["rolling_balance"] = true
+    # Try loading the demand flexibility parameters
+    demand_flexibility_parameters = DataFrames.DataFrame()
+    try
+        demand_flexibility_parameters = CSV.File(
+            joinpath(filepath, "demand_flexibility_parameters.csv")
+        ) |> DataFrames.DataFrame
+        println("...loading demand flexibility parameters")
 
-        # Try loading the demand flexibility parameters
-        demand_flexibility_parameters = DataFrames.DataFrame()
-        try
-            demand_flexibility_parameters = CSV.File(
-                joinpath(filepath, "demand_flexibility_parameters.csv")
-            ) |> DataFrames.DataFrame
-            println("...loading demand flexibility parameters")
-
-            # Create a dictionary to hold the error messages relevant to loading the 
-                # demand flexibility parameters
-            demand_flexibility_params_errs = Dict()
-            demand_flexibility_params_errs["duration"] = (
+        # Create a dictionary to hold the warning messages relevant to loading the 
+        # demand flexibility parameters
+        demand_flexibility_params_warns = Dict(
+            "duration" => (
                 "The demand flexibility duration parameter is not defined. Will "
-                * "default to nothing."
-            )
-            demand_flexibility_params_errs["enabled"] = (
+                * "default to being the size of the interval."
+            ),
+            "enabled" => (
                 "The parameter that indicates if demand flexibility is enabled is not "
                 * "defined. Will default to being enabled."
-            )
-            demand_flexibility_params_errs["interval_balance"] = (
+            ),
+            "interval_balance" => (
                 "The parameter that indicates if the interval load balance constraint "
                 * "is enabled is not defined. Will default to being enabled."
-            )
-            demand_flexibility_params_errs["rolling_balance"] = (
+            ),
+            "rolling_balance" => (
                 "The parameter that indicates if the rolling load balance constraint "
                 * "is enabled is not defined. Will default to being enabled."
-            )
+            ),
+        )
 
-            # Try assigning the different demand flexibility parameters from the file
-            for k in keys(demand_flexibility_params_errs)
-                try
-                    demand_flexibility[k] = demand_flexibility_parameters[1, k]
-                catch e
-                    println(demand_flexibility_params_errs[k])
-                end
-            end
-
-            # Set the demand flexibility constraints to false if enabled is false
-            if !demand_flexibility["enabled"]
-                demand_flexibility["interval_balance"] = false
-                demand_flexibility["rolling_balance"] = false
-            end
-        catch e
-            println("Demand flexibility parameters not found in " * filepath)
-            println(
-                "Demand flexibility parameters will default to allowing demand "
-                * "flexibility to occur."
-            )
-        end
-
-        # Check the feasibility of the duration parameter
-        if demand_flexibility["duration"] == nothing
-            demand_flexibility["duration"] = interval
-        elseif demand_flexibility["duration"] > interval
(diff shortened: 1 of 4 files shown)


lanesmith:
Mentioning @dongqi-wu so he can review this too

dongqi-wu:
> Mentioning @dongqi-wu so he can review this too
Hi Lane, I just read the code, they look quite straightforward and I don't have problems with them. Thanks!

src/read.jl
```diff
@@ -109,16 +109,33 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
-        demand_flexibility["flex_amt"] = nothing
+    else
+        if !isnothing(demand_flexibility["flex_amt_up"]) || (
+            !isnothing(demand_flexibility["flex_amt_dn"])
+        )
+            @warn (
+                "The exclusion of one of the demand flexibility profiles has resulted "
+                * "in demand flexibility not being enabled."
+            )
+        end
```
danielolsen: If we know that at least one of these is `nothing` from the `if` block above, do we need this `if` block here within the `else`?

danielolsen: Oh, I think I see now. This block is triggered if exactly one is present and exactly one is missing.

danielolsen (approved):
Looks good, thanks.

src/read.jl
lanesmith: Yeah, that was the intent! If both are `nothing`, then the user likely doesn't intend for demand flexibility to be present. However, if only one is `nothing`, the user may have forgotten to include one of the demand flexibility profiles. When one profile is excluded, demand flexibility won't work in the way that we have it set up, which is why we set `enabled` to `false` and include a more prominent warning to the user. If you would like, I can get rid of that `if` block and change the warning message to be `"The exclusion of at least one demand flexibility profile ..."`

danielolsen: I think the current behavior makes sense, it just took me a minute to parse the code properly in my head.

lanesmith:
Just to be sure, the plan was to merge this PR before #147, right?

BainanXia:
> Just to be sure, the plan was to merge this PR before #147, right?

I think the plan was merging #147 first. I will take a look at both this afternoon. Sorry for the delay.

danielolsen:
I think that this PR and #147 are mostly independent? If I cherry-pick the #147 commit on top of this one, there are a few merge conflicts since the same files are being edited, but they can be resolved fairly easily since the two PRs are doing different things. I assume the same will hold if we go the other way instead.

src/read.jl
```diff
@@ -192,6 +209,13 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
             )
             demand_flexibility["duration"] = interval
         end
+
+        # Prevent the rolling_balance constraint according to the duration parameter
+        if demand_flexibility["rolling_balance"]
+            if demand_flexibility["duration"] == interval
```
BainanXia: Just to confirm, I remember we project any user specified duration that is greater than the interval to the length of the interval, right?

src/read.jl
BainanXia: @lanesmith The logic makes sense to me. My questions is in the current setup, is `demand_flexibility` a parameter that the user could specify or it is still an internal indicator? If the user could set `demand_flexibility` parameter optionally, the current behavior could serve as the default behavior when it is not set explicitly by the user, i.e. `isnothing`. Otherwise, following situation still applies: the user would like to have demand flexibility, thus the parameter is set, however, none of the profiles are provided by accident.

lanesmith: I think what you are describing is the `enabled` parameter, which belongs to the `DemandFlexibility` struct and is specified in the `demand_flexibility_parameters.csv` file. So, yes, users can specify whether or not they want demand flexibility. However, if they don't have the requisite files (namely, the `demand_flexibility_up.csv` and `demand_flexibility_dn.csv` files), then demand flexibility cannot be enabled, regardless of the user's preference. There are a couple reasons I like checking for the demand flexibility profiles first, as opposed to checking for the `enabled` parameter (and, by extension, the `demand_flexibility_parameters.csv` file) first: (1) `demand_flexibility_profiles.csv` is not necessarily a required file meaning that, at a minimum, a user would only need to provide the `demand_flexibility_up.csv` and `demand_flexibility_dn.csv` files; and (2) if the profiles are not included, trying to read in the other files is not required. I'm happy to discuss this flow more if we would like to rearrange when things are read.

src/read.jl
danielolsen: Yes, see https://github.com/Breakthrough-Energy/REISE.jl/blob/lane/separate_demand_flex/src/read.jl#L203

lanesmith: Correct (I believe the code to which you are referring is listed directly above this snippet). I added this check because if the size of the rolling window is equal to the interval (at this point, the size of the rolling window cannot be larger than the interval due to the code about which you originally asked), then the `interval_balance` constraint and the `rolling_balance` constraint would be redundant.

src/read.jl
BainanXia: What I'm trying to say here is:

If the user specify `demand_flexibility_enabled = True`, but the profiles are not provided, we should throw an error instead of assuming `demand_flexibility_enabled = False` then running the simulation without it.

If the user specify `demand_flexibility_enabled = False`, but the profiles are provided, we ignore it. (I think it is the case in the current form)

If the user doesn't specify the parameter, i.e. `demand_flexibility_enabled == nothing`, the current logic makes perfect sense.

lanesmith: Okay, that makes sense. I'll make that change. In the first case that you mention, should it raise an error (causes the simulation to stop) or a warning (prints the warning, but the simulation continues using updated parameters set internally)?

BainanXia: I think if the user explicitly set `demand_flexibility_enabled = True` without providing the required profiles, we should raise an error and stop the simulation.

lanesmith: This should be done, though please look over `read_demand_flexibility` as it was refactored quite a bit. The way I was able to get the error message to be called is a little weird. When `throw`-ing an error, the resultant error at the top of the stacktrace is that `m` is not defined. In Julia, you can see the desired error text later in the stacktrace by just using `throw(ErrorException(...))`, however in Python you cannot (which kind of defeats the purpose of having the error message); apparently, Python does not produce as detailed of a stacktrace for the errors in the Julia code. So, that is why I have the main error message attached to `@error(...)`, which acts more like an error-centric print statement (it does not stop the simulation), and a minimal error message attached to `throw(ErrorException(...))`, which actually stops the simulation. If anybody knows of a better solution, I'd be happy to implement it

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
 end
 
 
-"""Load demand flexibility profile from .csv files into DataFrame(s)."""
+"""
+    read_demand_flexibility(filepath, interval)
```
BainanXia, line 108: Is it something we forget to delete here?

danielolsen: This is the docstring convention for Julia functions.

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
...
+
+Load demand flexibility profiles and parameters from .csv files and return them in a 
+DemandFlexibility struct.
```
BainanXia, line 111: I see the convention. Should we have an indent here?

lanesmith: I was going off of this: https://github.com/invenia/BlueStyle#documentation

danielolsen: It looks like the function call example should be indented, but the other documentation should not: https://docs.julialang.org/en/v1/manual/documentation/

BainanXia: I'm just comparing this with l.87 and l.45.

danielolsen: Aha, l.45 and l.87 of **model.jl**. I think in this case, both other style guides agree and so I think our style should be updated.

lanesmith: Ah, thanks for finding that Daniel, I wasn't sure what l.87 and l.45 were. I think as Daniel mentioned in a recent PR, we will probably need a more extensive PR at some point to refactor the `REISE.jl` code to be compliant with these style guides.

lanesmith: Would it be helpful for me to create an Issue focusing on refactoring the `REISE.jl` code to follow one of the other code styles? I know we discussed `Blue` in the past

danielolsen: Probably.

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
...
 function read_demand_flexibility(filepath, interval)::DemandFlexibility
     # Initialize demand flexibility
     demand_flexibility = Dict()
 
-    # Try loading demand flexibility profiles. Otherwise, create a DataFrame of zeros
+    # Pre-specify the demand flexibility parameters
+    demand_flexibility["enabled"] = true
+    demand_flexibility["duration"] = nothing
+    demand_flexibility["interval_balance"] = true
+    demand_flexibility["rolling_balance"] = true
```
BainanXia: Can we do this in one shot, something like
```julia
demand_flexibility = Dict([
   ("enabled", true),
   ("duration", nothing),
   ...
])
```
Also, should we set `demand_flexibility["enabled"] = "not specified"` instead of `true` according to the logic flow?

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
...
+        )
+        demand_flexibility_params_errs["interval_balance"] = (
+            "The parameter that indicates if the interval load balance constraint is "
+            * "enabled is not defined. Will default to being enabled."
+        )
+        demand_flexibility_params_errs["rolling_balance"] = (
+            "The parameter that indicates if the rolling load balance constraint is "
+            * "enabled is not defined. Will default to being enabled."
+        )
+
```
BainanXia: Same comment as above. Also, would be better to call it `warns` instead of `errs` following the convention that an error will stop the program?

src/read.jl
lanesmith: Yeah, I can do both of those things

src/read.jl
lanesmith: Yeah, I'll make those changes. Good call on the name change

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
...
-            demand_flexibility_params_errs["interval_balance"] = (
-                "The parameter that indicates if the interval load balance constraint "
-                * "is enabled is not defined. Will default to being enabled."
-            )
-            demand_flexibility_params_errs["rolling_balance"] = (
-                "The parameter that indicates if the rolling load balance constraint "
-                * "is enabled is not defined. Will default to being enabled."
-            )
+    # Check the feasibility of the duration parameter
+    if demand_flexibility["duration"] == nothing
```
BainanXia: Would it be simpler is we default this parameter to `interval` at the beginning?

lanesmith: Yeah, probably. That could remove the need for this `duration` check. I can make that change

src/read.jl
```diff
@@ -104,113 +104,160 @@ function read_storage(filepath)::Storage
...
-                catch e
-                    println(demand_flexibility_params_errs[k])
-                end
-            end
+    # Prevent the rolling_balance constraint according to the duration parameter
+    if demand_flexibility["rolling_balance"]
+        if demand_flexibility["duration"] == interval
+            demand_flexibility["rolling_balance"] = false
+        end
+    end
```
BainanXia: Would this work `demand_flexibility["rolling_balance"] &= ~(demand_flexibility["duration"]==interval)`?

lanesmith: Hmm not sure why I opted for the nested `if` statements in the first place. This is definitely much snazzier, thanks!

src/read.jl
lanesmith: Done.

src/read.jl
lanesmith: Done.

src/read.jl
lanesmith: Done.

src/read.jl
lanesmith: Done.

src/read.jl
```diff
@@ -104,113 +104,155 @@ function read_storage(filepath)::Storage
...
                 ) |> DataFrames.DataFrame
-                println("...loading demand flexibility $cost_file_suffix-shift cost profiles")
+                println("...loading demand flexibility " * s * "-shift cost profiles")
             catch e
                 println(
-                    "Demand flexibility down-shift cost profiles not found in " * filepath
+                    "Demand flexibility " 
+                    * s 
+                    * "-shift cost profiles not found in " 
+                    * filepath
```
BainanXia: In general the cost profiles are optional for demand flexibility to be considered. Probably, we could add another sentence here saying zero cost of elastic demand is considered?

BainanXia (approved):
This is nice. Thanks for the patience.

src/read.jl
lanesmith: Sure, I can add an additional sentence

lanesmith: Done.

PR #146, opened 2021-07-15
feat: allow extract-data to be specified through api

jenhagg:
### Purpose
Extend the flask api and underlying functions that are called in a container environment and locally, so that `--extract-data` can be passed to `call.py`, and the extraction can be done manually. This makes the interface consistent across server/container/local environments. Original motivation is to allow launching a bunch of scenarios in parallel, but serializing the extraction step to avoid race conditions. See https://github.com/Breakthrough-Energy/plug/issues/23. 

### What the code is doing
Some refactoring to reuse existing code, but mostly wiring things together.

### Testing
Created and launched a simulation in a container and on my local machine.  Also tested launching one with `scenario.launch_simulation(solver="clp", extract_data=False)` and got expected results.

### Time estimate
10 min

### DIFF
--- a/pyreisejl/utility/app.py
+++ b/pyreisejl/utility/app.py
@@ -11,44 +11,72 @@ app = Flask(__name__)
 
 
 """
-Example request:
+Example requests:
 
-curl -XPOST http://localhost:5000/launch/1234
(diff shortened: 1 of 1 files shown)


rouille (approved):
Nice

PR #147, opened 2021-07-16
feat: add demand flexibility cost

### LINKED ISSUE #124: Add cost components to demand flexibility model

dongqi-wu:

### Purpose
Update demand flexibility cost code in #135 to accommodate the changes in #142.  Consolidate comments from @lanesmith 

### What the code is doing
* The code attempts to read two additional input files `demand_flexibility_cost_up.csv` and `demand_flexibility_cost_up.csv` and create corresponding fields in the DemandFlexibility class
* The numeric costs are added to the decision variables `load_shift_up` and `load_shift_dn`

### Testing
The code is setup on a local computer and uses a ERCOT sub-system to reduce computation time
The code is tested using dummy input files (shared in slack channel) and when inputs files are missing
The results with/without the cost input files are checked and compared
Various testing are performed using interval length and interval number ranging from 1 to 3 

### Where to look
* It's helpful to clarify where your new code lives if you moved files around or there could be confusion/

* two fields in types.jl
* csv reading code in read.jl
* add corresponding objective function expression in model.jl
* re-assign coefficients in loop.jl

### Usage Example/Visuals
N/A

### Time estimate
~15 min for people that are familiar with the previous version in #135

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,12 +29,14 @@ RUN mkdir src && touch src/REISE.jl
 RUN julia -e 'using Pkg; \
 	Pkg.activate("."); \
 	Pkg.instantiate(); \
+	Pkg.add("GLPK"); \
 	Pkg.add("Gurobi"); \
-	Pkg.add("GLPK")'
+	Pkg.add("Clp")'
 
 COPY src src
 
 RUN julia -e 'import Gurobi; \
+	import Clp; \
 	import GLPK; \
 	using REISE'
 


src/model.jl
```diff
@@ -579,10 +601,26 @@ function _build_model(
...
+                )
+            )
+        end
+        if !isnothing(demand_flexibility.cost_dn)
+            bus_demand_flex_cost_dn = permutedims(
+                Matrix(
+                    demand_flexibility.cost_dn[start_index:end_index, 2:end]
+                )
+            )
+        end
```
danielolsen: Can we move these blocks within the `_add_objective_function!` function? The reason that we calculate `bus_demand_flex_amt` within `_build_model` is because we need it to set bounds for the `load_shift_up` and `load_shift_up` variables, but I believe we can continue to pass only `demand_flexibility` to `_add_objective_function!` and handle the flexibility costs within there.

src/loop.jl
```diff
@@ -57,6 +57,20 @@ function interval_loop(factory_like, model_kwargs::Dict,
...
+                )
+            )
+        end
+        if !isnothing(demand_flexibility.cost_dn)
+            bus_demand_flex_cost_dn = permutedims(
+                Matrix(
+                    demand_flexibility.cost_dn[interval_start:interval_end, 2:end]
+                )
+            )
+        end
```
danielolsen: Can we move these within the `if demand_flexibility.enabled` block below?

src/model.jl
lanesmith: Agreed. That would also prevent `bus_demand_flex_cost_up` and `bus_demand_flex_cost_dn` from needing to be initialized as `nothing` a few lines above.

src/read.jl
```diff
@@ -192,6 +194,32 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+            demand_flexibility["cost_up"] = CSV.File(
+                joinpath(filepath, "demand_flexibility_cost_up.csv")
+            ) |> DataFrames.DataFrame
+            println("...loading demand flexibility up-shift cost profiles")
+        catch e
+            println(
+                "Demand flexibility up-shift cost profiles not found in " * filepath
+            )
+            demand_flexibility["cost_up"] = nothing
+        end
```
lanesmith: Can we reduce the redundancy between these two blocks? Since they both have the same structure, we can probably use a `for` loop to iterate through `["up", "dn"]`. If we want to take this approach, we can mimic what is implemented in #145 for reading the up and down demand flexibility profiles.

lanesmith:
Aside from the comments that have been left, this looks good to me. I'm glad to see that using `set_objective_coefficient` ended up working out in `loop.jl`. I'm not sure what it was, but when I tested the system last week (with the old laptop and old environment), I was still seeing the same error that we previously discussed. When I tested it this morning with the new environment, it worked! Maybe I was using an older `JuMP` version or something (though I'm not sure if that would be the case because the `Project.toml` file was up-to-date).

src/model.jl
dongqi-wu: @danielolsen @lanesmith Hi, I just saw the comments, I will address them and update this PR soon.

PR #153, opened 2021-07-28
fix: define interval length for load shed constraint function

danielolsen:

### Purpose
Fix a bug introduced in #142, in which the `interval_length` variable is not defined within the `_add_constraint_load_shed!` function.

### What the code is doing
Defining the variable, based on the dimensions of the existing bus_demand input.

### Testing
In progress.

### Time estimate
2 minutes.


BainanXia (approved):
Thanks.

PR #154, opened 2021-08-04
Pass net demand deviations to the next interval

### LINKED ISSUE #123: Pass net demand deviations to the next interval
Currently, net demand deviations (caused by demand flexibility) at the end of one interval are not passed to the next interval. With the way the load balance constraints are implemented, a nonnegative amount of net shifted demand can be available at the end of an interval (though it probably will not occur frequently since the cost minimization is seeking to curtail demand, not add more). If more demand is added in an interval, it makes sense that the additional demand would no longer need to be met in the next interval. Similar to the way an interval's final state of charge is passed to the next interval's initial state of charge in the battery energy storage model, an interval's final net demand deviation could be passed to the next interval's initial net demand deviation.

An interval's final net demand deviation would likely be equal to `sum(load_shift_up[t] - load_shift_dn[t])`, where `t` equals all time steps in the interval (this would be different if a look-ahead period were to be added; you would not want to include net demand deviations that occur during the look-ahead period). An interval's initial net demand deviation (which is set equal to the prior interval's final net demand deviation) would modify the interval load balance constraint and the first rolling window of the rolling load balance constraint.

This functionality will likely become even more important if/when a look-ahead period is added. If a look-ahead period is added, the interval load balance constraint would not necessarily make the net demand deviation at the end of an interval close to (or slightly greater than) zero. Both nonnegative and nonpositive net demand deviations could be passed to the next interval depending on what is necessitated by the look-ahead period. Implementation of the look-ahead period will likely be a bit more involved, as this functionality will touch other resources (e.g., battery energy storage) and will increase computation time (this creates other issues, such as determining look-ahead periods of meaningful length).

lanesmith:

### Purpose
The purpose of this feature is to pass net demand deviations from the current interval to the following interval. Based on the demand flexibility formulation, shifted demand does not necessarily need to be completely balanced at the end of an interval (though it likely will be due to the cost minimizing objective). This feature ensures that all shifted demand ends up being accounted for. This PR closes #123.

### What the code is doing
In `model.jl`, a new constraint (`rolling_load_balance_first`) is added to account for the previous interval's net demand deviation during instances where the `rolling_balance` parameter is enabled. This new constraint creates an additional rolling window at the beginning of the interval to balance demand deviations and the previous interval's net demand deviation. Additionally, the `interval_load_balance` constraint is updated to account for the net demand deviation. There is also a small fix to the number of rolling windows that are used in the `rolling_load_balance` constraint. Previous iterations of the model did not accurately account for the number of balancing periods that should be present.

In `loop.jl`, the net demand deviation is calculated and passed to the following interval. This implementation is similar to that of the initial state of charge parameter used in the energy storage model. Net demand deviations are calculated as the difference between the up and down deviations for each bus with demand flexibility for each interval. The `interval_load_balance` and `rolling_load_balance_first` constraints are also updated appropriately for intervals `i > 2`.

### Testing
I tested this on the iteration of the ERCOT grid that we have been using for testing. When testing on 3 intervals of 24 hours each, net demand deviations are passed to the following interval without issue.

### Where to look
Changes are contained within `model.jl` and `loop.jl`.

### Time estimate
This hopefully shouldn't take too long to review. A good chunk of the noted changes are due to how the demand flexibility constraints and variables are updated in `loop.jl`. No functionality is changed, just how they are set in the for loops.

### DIFF
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -66,7 +66,15 @@ function interval_loop(factory_like, model_kwargs::Dict,
                 model_kwargs["storage_e0"] = storage.sd_table.InitialStorage
             end
             if demand_flexibility.enabled
-                model_kwargs["init_shifted_demand"] = zeros(size(bus_flex_amt, 1))
+                try
+                    model_kwargs["init_shifted_demand"] = zeros(
+                        size(bus_demand_flex_amt_dn, 1)
+                    )
+                catch e
+                    model_kwargs["init_shifted_demand"] = zeros(
+                        size(bus_demand_flex_amt_up, 1)
+                    )
+                end
             end
             m = new_model(factory_like)
             JuMP.set_optimizer_attributes(m, pairs(solver_kwargs)...)
@@ -174,12 +182,17 @@ function interval_loop(factory_like, model_kwargs::Dict,
                             )
                         end
                     end
-                    JuMP.set_normalized_rhs(
-                        m[:rolling_load_balance_first][l], -1 * init_shifted_demand[l]
-                    )
-                    JuMP.set_normalized_rhs(
-                        m[:interval_load_balance][l], -1 * init_shifted_demand[l]
-                    )
+                    if demand_flexibility.rolling_balance
+                        JuMP.set_normalized_rhs(
+                            m[:rolling_load_balance_first][l], 
+                            -1 * init_shifted_demand[l],
+                        )
+                    end
+                    if demand_flexibility.interval_balance
+                        JuMP.set_normalized_rhs(
+                            m[:interval_load_balance][l], -1 * init_shifted_demand[l]
+                        )
+                    end
                 end
             end
         end
@@ -257,18 +270,7 @@ function interval_loop(factory_like, model_kwargs::Dict,
             storage_e0 = results.storage_e[:,end]
         end
         if demand_flexibility.enabled
-            if demand_flexibility.rolling_balance && (
-                demand_flexibility.duration < interval
-            )
-                init_shifted_demand = sum(
-                    results.load_shift_up[
-                        :, (interval - demand_flexibility.duration + 1):end
-                    ] - results.load_shift_dn[
-                        :, (interval - demand_flexibility.duration + 1):end
-                    ],
-                    dims=2
-                )[:, 1]
-            else
+            if demand_flexibility.interval_balance || demand_flexibility.rolling_balance
                 init_shifted_demand = sum(
                     results.load_shift_up - results.load_shift_dn,
                     dims=2
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -66,15 +66,9 @@ function interval_loop(factory_like, model_kwargs::Dict,
                 model_kwargs["storage_e0"] = storage.sd_table.InitialStorage
             end
             if demand_flexibility.enabled
-                try
-                    model_kwargs["init_shifted_demand"] = zeros(
-                        size(bus_demand_flex_amt_dn, 1)
-                    )
-                catch e
-                    model_kwargs["init_shifted_demand"] = zeros(
-                        size(bus_demand_flex_amt_up, 1)
-                    )
-                end
+                model_kwargs["init_shifted_demand"] = zeros(
+                    size(bus_demand_flex_amt_dn, 1)
+                )
             end
             m = new_model(factory_like)
             JuMP.set_optimizer_attributes(m, pairs(solver_kwargs)...)
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -265,10 +265,13 @@ function interval_loop(factory_like, model_kwargs::Dict,
         end
         if demand_flexibility.enabled
             if demand_flexibility.interval_balance || demand_flexibility.rolling_balance
-                init_shifted_demand = sum(
-                    results.load_shift_up - results.load_shift_dn,
-                    dims=2
-                )[:, 1]
+                init_shifted_demand = dropdims(
+                    sum(results.load_shift_up - results.load_shift_dn, dims=2); dims=2
+                )
+            else
+                init_shifted_demand = zeros(
+                    size(bus_demand_flex_amt_dn, 1)
+                )
             end
         end
 
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -66,7 +66,7 @@ function interval_loop(factory_like, model_kwargs::Dict,
                 model_kwargs["storage_e0"] = storage.sd_table.InitialStorage
             end
             if demand_flexibility.enabled
-                model_kwargs["init_shifted_demand"] = zeros(size(bus_flex_amt, 2))
+                model_kwargs["init_shifted_demand"] = zeros(size(bus_flex_amt, 1))
             end
             m = new_model(factory_like)
             JuMP.set_optimizer_attributes(m, pairs(solver_kwargs)...)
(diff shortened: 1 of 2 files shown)


src/loop.jl
```diff
@@ -65,6 +65,17 @@ function interval_loop(factory_like, model_kwargs::Dict,
...
+            if demand_flexibility.enabled
+                try
+                    model_kwargs["init_shifted_demand"] = zeros(
+                        size(bus_demand_flex_amt_dn, 1)
+                    )
+                catch e
+                    model_kwargs["init_shifted_demand"] = zeros(
+                        size(bus_demand_flex_amt_up, 1)
+                    )
+                end
```
danielolsen: Which exceptions are we expecting to be caught here?

lanesmith: I added this in case a profile in one direction was not provided. However, this shouldn't be a problem based on the functionality added to `read_demand_flexibility` in `read.jl`. I can remove the `try`-`catch` check

lanesmith: Done.

src/model.jl
```diff
@@ -332,18 +332,28 @@ function _add_constraints_demand_flexibility!(
...
+                for j in 1:(demand_flexibility.duration - 1)
+            ) >= -1 * init_shifted_demand[i],
+        )
         println("rolling load balance: ", Dates.now())
         JuMP.@constraint(
             m,
             rolling_load_balance[
                 i in 1:sets.num_load_bus,
-                k in 1:(interval_length - demand_flexibility.duration)
+                k in 1:(interval_length - demand_flexibility.duration + 1)
```
BainanXia, line 383: Nice catch.

src/loop.jl
```diff
@@ -75,6 +80,9 @@ function interval_loop(factory_like, model_kwargs::Dict,
             if storage.enabled
                 model_kwargs["storage_e0"] = storage_e0
             end
+            if demand_flexibility.enabled
+                model_kwargs["init_shifted_demand"] = init_shifted_demand
+            end
```
BainanXia, line 85: Here we are assuming implicitly if demand flexibility is enabled, at least one of interval balance and rolling balance is enabled?

src/loop.jl
```diff
@@ -128,35 +136,55 @@ function interval_loop(factory_like, model_kwargs::Dict,
...
                     bus_demand_flex_cost_dn = permutedims(
                         Matrix(
-                            demand_flexibility.cost_dn[interval_start:interval_end, 2:end]
+                            demand_flexibility.cost_dn[
+                                interval_start:interval_end, 2:end
+                            ]
                         )
                     )
                 end
-                for t in 1:interval, i in 1:length(sets.load_bus_idx)
```
BainanXia: Aha, previously we used same variable `i` as the outer loop variable here, which will cause dead loop, right?

src/loop.jl
```diff
@@ -235,6 +263,14 @@ function interval_loop(factory_like, model_kwargs::Dict,
         if storage.enabled
             storage_e0 = results.storage_e[:,end]
         end
+        if demand_flexibility.enabled
+            if demand_flexibility.interval_balance || demand_flexibility.rolling_balance
+                init_shifted_demand = sum(
+                    results.load_shift_up - results.load_shift_dn,
+                    dims=2
+                )[:, 1]
```
BainanXia: We are taking the first column from the summation here, what's in the rest columns? I presume it should be one column in total, no?

src/loop.jl
lanesmith: Not in Julia. Just doing some basic testing, it appears that it handles this problem pretty well. If both an inner loop and outer loop use `i`, then the outer loop will pick back up after the inner loop completes. There would be an issue if you wanted to use both the outer loop and inner loop variables. There wasn't a problem here, I just changed it for readability (and correctness) purposes.

src/loop.jl
lanesmith: Ah good catch. This probably should not be an implicit assumption since the user could choose to not have any of the demand recovered (I guess this is one way to have sheddable demand enabled), i.e. both interval balance and rolling balance disabled. In that case, we would not want net demand deviations passed. I'll update this `if` statement

src/loop.jl
BainanXia: Thanks for the explanation. That's good to know.

src/loop.jl
lanesmith: Yes you are right. I'm trying to remember the exact issue, but I believe what was happening was that a two-dimensional array was being returned when we really just want a one-dimensional array; mathematically they were the same, but code-wise they were not. Using `[:, 1]` was just a way to "reshape" the array from two dimensional to one dimensional.

danielolsen: I think the `dropdims` function may be a clearer way to do this.

lanesmith: Thanks for letting me know! I'll check it out

src/model.jl
```diff
@@ -332,18 +332,28 @@ function _add_constraints_demand_flexibility!(
...
+    init_shifted_demand::Array{Float64,1}=Float64[],
 )
     if demand_flexibility.rolling_balance
+        println("rolling load balance, first window: ", Dates.now())
+        JuMP.@constraint(
+            m,
+            rolling_load_balance_first[i in 1:sets.num_load_bus],
+            sum(
+                m[:load_shift_up][i, j] - m[:load_shift_dn][i, j] 
+                for j in 1:(demand_flexibility.duration - 1)
```
BainanXia, line 375: Should `j` be ranged in `[1, demand_flexibility.duration]` here?

src/model.jl
BainanXia: Should `k` start from 2 here?

src/model.jl
lanesmith: I don't think so. The idea for incorporating the previous interval's net demand deviation to the first rolling balance window is that the net demand deviation is like time `j = 0`. So, if we have `demand_flexibility.duration = 6`, then the first window would include 0 (the previous interval's net demand deviation) and 1, 2, 3, 4, and 5 (`1:(demand_flexibility.duration - 1)`)

src/model.jl
lanesmith: I think `k` should start at 1 here. I'm treating the net demand deviation as time `k = 0`, so the first rolling balance window (the separate constraint) accounts for that. Then the second window, which is included in this constraint, would start at `k = 1`

src/model.jl
BainanXia: Should it be interpreted as "demand deviation of the first duration window + leftovers from previous interval >= 0"? Here, `j` should loop through every hour in the first duration window, no?

lanesmith: Yeah, that's pretty much what I was trying to go for. Except, the first duration window is one hour shorter than normal (`(demand_flexibility.duration - 1)`) because the "leftovers from previous interval" are essentially take up one time step

BainanXia: I see. Is that the classic way to model this? It seems a little weird to me since in the very first duration, i.e. first duration at the first interval, this gives the load balancing among hours `demand_flexibility.duration - 1`. Also, we didn't follow the same logic in interval balance either, i.e. `init_shifted_demand` is added into the constraint over the entire interval instead of excluding the last hour of the interval?

src/loop.jl
lanesmith: Done.

src/model.jl
BainanXia: Yeah, I see this is related to our discussion above.

src/loop.jl
lanesmith: Done. I had actually kind of accounted for this on line 267 of `loop.jl`. I  also added an `else` statement there to catch the times that both interval balance and rolling balance are not implemented.

BainanXia (approved):
An agreement has been reached on the current implementation via an offline discussion. Thanks!

PR #155, opened 2021-08-11
feat: allow optional bus-level demand flexibility and cost

### LINKED ISSUE #140: Increase the Granularity of Demand Flexibility Input File
# :rocket:

- [x] Is your feature request essential for your project?

I wish to increase the data granularity of the input file `demand_flexibility.csv`. The current input assigns a number per load-zone which get later converted and applied to all load buses in the that load-zone in the opt problem formulation. This implies all loads in each load-zone have the same flexibility which is not a very realistic representation flexible demand behaviour.  I suggest instead of assigning ONE number to all buses in a load-zone, the user should be able to specify the flexibility of each load bus individually.

This should provide a description of the feature request, e.g.:
* The dimension of input file `demand_flexibility.csv` should be (hour_num, load_bus_num) instead of (hour_num, load_zone_num)
* The function `_make_bus_demand_flexibility_amount` should be adjusted appropriately such that it selects a partition of `demand_flexibility.flex_amt` that corresponds to the interval of the model, instead of creating a matrix using the function `permutedims()`.

### Describe alternatives you've considered, if relevant
An alternative or improved implementation is proposed in issue #141

### LINKED ISSUE #141: Add an User Option that Allows Users to Specify the Form of  Demand Flexibility Input Data
# :rocket:

- [x] Is your feature request essential for your project?

I wish the user can choose to provide the demand flexibility input data in several different formats and tell the program to interpret through an additional user-specified parameter in `demand_flexibility_parameters.csv`. Currently, the input csv file `demand_flexibility.csv` has the resolution of per-load-zone, which means the program assigns a single number to all loads in every load zone. Issue #140 proposed an improvement to the input data granularity. 

The current implementation add two demand flexibility constraints for each load bus in the case if data is provided. However, in demand response simulations one may be only interested to consider a small subset of load buses in the case that represents aggregated urban or industrial loads. With the current implementation it is impossible to alter the flexibility of individual buses without changing all other load buses in the same load zone. Additionally, if an user just want to study the impact of demand flexibility on a subset of load buses, adding the constraints for the buses that are not assigned to be flexible also introduce unnecessary computation complexity.

* The input file `demand_flexibility_parameters.csv` should have one additional parameter that allows the user to choose from two (more if needed) modes: load-zone level or bus level. For load-zone level input, the user specifies a number for each load zone and the number of columns of `demand_flexibility_parameters.csv` should be exactly equal to the number of load-zones. For bus-level input, the user can specify the flexibility of any number of buses in the system, the number of columns in `demand_flexibility.csv` should be within the range [0, load_bus_num].
* The class `demand_flexibility` should have an additional field corresponding to the new parameter.
* In the data pre-processing in `model.jl`, there function `_make_bus_demand_flexibility_amount()` should check for the new parameter and compute the current demand flexibility matrix. 
* The dimension of decision variables `load_shift_up` and `load_shift_dn` should be equal to the number of load buses that have a non-zero flexibility, instead of all load buses. The parts that place constraints on the two variables should also be changed accordingly.

### Describe alternatives you've considered, if relevant
This proposal is based on issue #140.

dongqi-wu:

### Purpose
Implement functionalities related to issue #141

### What the code is doing
Compare to the main branch, this version does the following:
* read a new string variable that is either "AREA" or "BUS" from the parameter csv
* if "AREA", do exactly the same thing as the active main branch
* if "BUS", the code look for the buses (columns in input demand_flexibility_up/dn.csv) and create decision variables and constraints only for the buses with a specify flexibility

### Testing
Testing both functions and the entire REISE program using dummy input
The numerical result coming out of the program is not yet analyzed
 
### Where to look
* type.jl -- definition of Sets and DemandFlexibility
* loop.jl -- minor change
* read.jl -- read one additional parameter from input csv
* model.jl -- many functions changed to accommodate the new functionality

### Time estimate
~45 min

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -91,23 +91,29 @@ function _make_bus_demand_flexibility_amount(
 )::Tuple{Matrix,Matrix}
     # Bus weighting
     zone_to_bus_shares = _make_bus_demand_weighting(case)
-
     # Demand flexibility up profile
     simulation_demand_flex_amt_up = Matrix(
         demand_flexibility.flex_amt_up[start_index:end_index, 2:end]
     )
-    bus_demand_flex_amt_up = permutedims(
-        simulation_demand_flex_amt_up * zone_to_bus_shares
-    )
-
     # Demand flexibility down profile
     simulation_demand_flex_amt_dn = Matrix(
         demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
     )
-    bus_demand_flex_amt_dn = permutedims(
-        simulation_demand_flex_amt_dn * zone_to_bus_shares
-    )
-
+    # convert per-area input to per-bus values
+    if demand_flexibility.input_granularity == "AREA"
+        bus_demand_flex_amt_up = permutedims(
+            simulation_demand_flex_amt_up * zone_to_bus_shares
+        )
+        bus_demand_flex_amt_dn = permutedims(
+            simulation_demand_flex_amt_dn * zone_to_bus_shares
+        )
+    # input is already per-bus, copy to new container
+    elseif demand_flexibility.input_granularity == "BUS"
+        bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
+        bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
+    else
+        println("Warning! Invalid demand flexibility parameter input_granularity!")
+    end
     return (bus_demand_flex_amt_up, bus_demand_flex_amt_dn)
 end
 
@@ -133,12 +139,10 @@ function _build_segment_slope(case::Case, segment_idx, segment_width)::Matrix
 end
 
 
-function _make_sets(case::Case)::Sets
-    _make_sets(case, nothing)
-end
-
-
-function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
+function _make_sets(case::Case, 
+                    storage::Union{Storage,Nothing}=nothing, 
+                    demand_flexibility::Union{DemandFlexibility, Nothing}=nothing
+        )::Sets
     # Positional indices from mpc.gencost
     MODEL = 1
     NCOST = 4
@@ -180,11 +184,31 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
                                * "Did you forget to linearize_gencost?")
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
+    # Demand flexibility, additional info for input->bus conversion
+    if demand_flexibility.enabled
+        if demand_flexibility.input_granularity == "BUS"
+            # assume each flexible bus can go up/dn, so only use the Dataframe for up
+            flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
+            flexible_bus_id = [parse(Int64, bus) for bus in flexible_bus_str]
+            flexible_bus_idx = [bus_id2idx[bus] for bus in flexible_bus_id]
+            num_flexible_bus = length(flexible_bus_idx)
+            flexible_load_bus_map = sparse(
+                    flexible_bus_idx, 1:num_flexible_bus, 1, num_bus, num_flexible_bus
+                )::SparseMatrixCSC
+        elseif demand_flexibility.input_granularity == "AREA"
+            flexible_bus_idx = load_bus_idx
+            num_flexible_bus = num_load_bus
+            flexible_load_bus_map = load_bus_map
+        end
+    else
+        flexible_bus_idx = nothing
+        num_flexible_bus = 0
+        flexible_load_bus_map = nothing
+    end
     # Storage
     storage_enabled = isa(storage, Storage) && storage.enabled
     num_storage = storage_enabled ? size(storage.gen, 1) : 0
     storage_idx = storage_enabled ? (1:num_storage) : nothing
-
     sets = Sets(;
         num_bus=num_bus, bus_idx=bus_idx, bus_id2idx=bus_id2idx,
         load_bus_idx=load_bus_idx, num_load_bus=num_load_bus, load_bus_map=load_bus_map,
@@ -196,7 +220,9 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
         gen_wind_idx=gen_wind_idx, renewable_idx=renewable_idx,
         num_wind=num_wind, num_solar=num_solar, num_hydro=num_hydro,
         num_segments=num_segments, segment_idx=segment_idx,
-        num_storage=num_storage, storage_idx=storage_idx)
+        num_storage=num_storage, storage_idx=storage_idx, 
+        flexible_bus_idx=flexible_bus_idx, num_flexible_bus=num_flexible_bus,
+        flexible_load_bus_map=flexible_load_bus_map)     
     return sets
 end
 
@@ -232,10 +258,10 @@ function _add_constraint_power_balance!(
     end
     if demand_flexibility.enabled
         injections = JuMP.@expression(
-            m, injections + sets.load_bus_map * m[:load_shift_dn]
+            m, injections + sets.flexible_load_bus_map * m[:load_shift_dn]
         )
         withdrawals = JuMP.@expression(
-            m, withdrawals + sets.load_bus_map * m[:load_shift_up]
+            m, withdrawals + sets.flexible_load_bus_map * m[:load_shift_up]
         )
     end
     JuMP.@constraint(m, powerbalance, (injections .== withdrawals))
@@ -263,10 +289,15 @@ function _add_constraint_load_shed!(
         bus_demand[sets.load_bus_idx[i], j],
     )
     if demand_flexibility.enabled
+        flexible_load_bus_idx = indexin(sets.flexible_bus_idx, sets.load_bus_idx)
+        flexible_load_map = sparse(
+                flexible_load_bus_idx, 1:sets.num_flexible_bus, 1, sets.num_load_bus, sets.num_flexible_bus
+            )::SparseMatrixCSC
         demand_for_load_shed = JuMP.@expression(
-            m,
-            [i=1:sets.num_load_bus, j=1:interval_length],
-            demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
+            m, 
+            demand_for_load_shed
+            + flexible_load_map * m[:load_shift_up]
+            - flexible_load_map * m[:load_shift_dn],
         )
     end
     JuMP.@constraint(
@@ -338,7 +369,7 @@ function _add_constraints_demand_flexibility!(
         JuMP.@constraint(
             m,
             rolling_load_balance[
-                i in 1:sets.num_load_bus,
+                i in 1:sets.num_flexible_bus,
                 k in 1:(interval_length - demand_flexibility.duration)
             ],
             sum(
@@ -351,7 +382,7 @@ function _add_constraints_demand_flexibility!(
         println("interval load balance: ", Dates.now())
         JuMP.@constraint(
             m,
-            interval_load_balance[i in 1:sets.num_load_bus],
+            interval_load_balance[i in 1:sets.num_flexible_bus],
             sum(
                 m[:load_shift_up][i, j] - m[:load_shift_dn][i, j] for j in 1:interval_length
             ) >= 0,
@@ -577,7 +608,6 @@ function _add_objective_function!(
             JuMP.add_to_expression!(obj, demand_response_penalty_dn)
         end
     end
-    
     # Finally, set as objective of model
     JuMP.@objective(m, Min, obj)
 end
@@ -609,14 +639,12 @@ function _build_model(
     # Positional indices from mpc.gen
     PMAX = 9
     PMIN = 10
-
     println("building sets: ", Dates.now())
     # Sets - time periods
     hour_idx = 1:interval_length
     end_index = start_index + interval_length - 1
     # Sets - static
-    sets = _make_sets(case, storage)
-
+    sets = _make_sets(case, storage, demand_flexibility)
     println("parameters: ", Dates.now())
     # Parameters
     bus_demand = _make_bus_demand(case, start_index, end_index) * demand_scaling
@@ -669,16 +697,15 @@ function _build_model(
     if demand_flexibility.enabled
         # The amount of demand that is curtailed from the base load
         JuMP.@variable(
-            m, 
-            0 <= load_shift_dn[i in 1:sets.num_load_bus, j in 1:interval_length] 
-                <= bus_demand_flex_amt_dn[sets.load_bus_idx[i], j],
+            m,
+            0 <= load_shift_dn[i in 1:sets.num_flexible_bus, j in 1:interval_length]
+                <= bus_demand_flex_amt_dn[i, j],
         )
-
-        # The amount of demand that is added to the base load
+        # The amount of demand that is added from the base load
         JuMP.@variable(
-            m, 
-            0 <= load_shift_up[i in 1:sets.num_load_bus, j in 1:interval_length]
-                <= bus_demand_flex_amt_up[sets.load_bus_idx[i], j],
+            m,
+            0 <= load_shift_up[i in 1:sets.num_flexible_bus, j in 1:interval_length]
+                <= bus_demand_flex_amt_up[i, j],
         )
     end
 
--- a/src/types.jl
+++ b/src/types.jl
@@ -52,6 +52,7 @@ Base.@kwdef struct DemandFlexibility
     enabled::Bool
     interval_balance::Bool
     rolling_balance::Bool
+    input_granularity::String
 end
 
 
@@ -102,6 +103,10 @@ Base.@kwdef struct Sets
     # Segments
     num_segments::Int64
     segment_idx::UnitRange{Int64}
+    ## Demand Flexibility
+    flexible_bus_idx::Union{Array{Int64,1}, Nothing}
+    num_flexible_bus::Int64
+    flexible_load_bus_map::Union{SparseMatrixCSC, Nothing}
     # Storage
     num_storage::Int64
     storage_idx::Union{UnitRange{Int64},Nothing}
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -41,7 +41,7 @@ function interval_loop(factory_like, model_kwargs::Dict,
     case = model_kwargs["case"]
     storage = model_kwargs["storage"]
     demand_flexibility = model_kwargs["demand_flexibility"]
-    sets = _make_sets(case, storage)
+    sets = _make_sets(case, storage, demand_flexibility)
     unused_load_shed_intervals_turnoff = 14
     # Start looping
     for i in 1:n_interval
@@ -139,14 +139,14 @@ function interval_loop(factory_like, model_kwargs::Dict,
                         )
                     )
                 end
-                for t in 1:interval, i in 1:length(sets.load_bus_idx)
+                for t in 1:interval, i in 1:sets.num_flexible_bus
                     JuMP.set_upper_bound(
                         m[:load_shift_up][i, t], 
-                        bus_demand_flex_amt_up[sets.load_bus_idx[i], t],
+                        bus_demand_flex_amt_up[i, t],
                     )
                     JuMP.set_upper_bound(
                         m[:load_shift_dn][i, t], 
-                        bus_demand_flex_amt_dn[sets.load_bus_idx[i], t],
+                        bus_demand_flex_amt_dn[i, t],
                     )
                     
                     if !isnothing(demand_flexibility.cost_up)
@@ -172,13 +172,13 @@ function interval_loop(factory_like, model_kwargs::Dict,
             status = JuMP.termination_status(m)
             if status == JuMP.MOI.OPTIMAL
                 f = JuMP.objective_value(m)
-                results = get_results(f, model_kwargs["case"])
+                results = get_results(f, model_kwargs["case"], demand_flexibility)
                 break
             elseif ((status == JuMP.MOI.LOCALLY_SOLVED)
                     & ("load_shed_enabled" in keys(model_kwargs)))
                 # if load shedding is enabled, we'll accept 'suboptimal'
                 f = JuMP.objective_value(m)
-                results = get_results(f, model_kwargs["case"])
+                results = get_results(f, model_kwargs["case"], demand_flexibility)
                 break
             elseif ((status in numeric_statuses)
                     & (JuMP.solver_name(m) == "Gurobi")
--- a/src/query.jl
+++ b/src/query.jl
@@ -3,9 +3,9 @@
 
 Extract the results of a simulation, store in a struct.
 """
-function get_results(f::Float64, case::Case)::Results
+function get_results(f::Float64, case::Case, demand_flexibility::DemandFlexibility)::Results
     status = "OPTIMAL"
-    sets = _make_sets(case)
+    sets = _make_sets(case, nothing, demand_flexibility)
     # These variables will always be in the results
     pg = JuMP.value.(m[:pg])
     pf = JuMP.value.(m[:pf])
@@ -69,8 +69,8 @@ function get_results(f::Float64, case::Case)::Results
     try
         load_shift_up_temp = JuMP.value.(m[:load_shift_up])
         load_shift_dn_temp = JuMP.value.(m[:load_shift_dn])
-        load_shift_up = sets.load_bus_map * load_shift_up_temp
-        load_shift_dn = sets.load_bus_map * load_shift_dn_temp
+        load_shift_up = sets.flexible_load_bus_map * load_shift_up_temp
+        load_shift_dn = sets.flexible_load_bus_map * load_shift_dn_temp
     catch e
         if isa(e, KeyError)
             # Thrown when load shift variables are `nothing`
--- a/src/read.jl
+++ b/src/read.jl
@@ -117,6 +117,7 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         "enabled" => "not_specified",  
         "interval_balance" => true, 
         "rolling_balance" => true,
+        "input_granularity" => "AREA",
     )
 
     # Try loading the demand flexibility parameters
@@ -145,6 +146,11 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
             "rolling_balance" => (
                 "The parameter that indicates if the rolling load balance constraint "
                 * "is enabled is not defined. Will default to being enabled."
+            ),
+            "input_granularity" => (
+                "The parameter that indicates the way demand flexibility "
+                * "and demand flexibility cost are provided is not defined. "
+                * "Will default to being area-based."
             ),
         )
 


src/REISE.jl
```diff
@@ -76,11 +76,11 @@ function run_scenario(;
...
     # While redirecting stdout and stderr...
     println("Redirecting outputs, see stdout.log & stderr.err in outputfolder")
-    redirect_stdout_stderr(stdout_filepath, stderr_filepath) do
+    #redirect_stdout_stderr(stdout_filepath, stderr_filepath) do
         # Loop through intervals
-        m = interval_loop(optimizer_factory, model_kwargs, solver_kwargs, interval,
+    m = interval_loop(optimizer_factory, model_kwargs, solver_kwargs, interval,
                           n_interval, start_index, inputfolder, outputfolder)
-    end
+    #end
```
danielolsen: If the code is working, should we still be exiting the stdout redirection here?

src/model.jl
```diff
@@ -609,14 +676,13 @@ function _build_model(
...
     PMIN = 10
-
     println("building sets: ", Dates.now())
     # Sets - time periods
     hour_idx = 1:interval_length
     end_index = start_index + interval_length - 1
     # Sets - static
-    sets = _make_sets(case, storage)
-
+    print(123)
```
danielolsen: This is probably test code that can be removed?

src/model.jl
```diff
@@ -609,14 +676,13 @@ function _build_model(
...
+    sets = _make_sets(case, storage, demand_flexibility)
```
danielolsen, line 647: You mentioned on the call that there was some problem you encountered with the `_make_sets` function. Could you please tell us a little more about what you were trying to do, and what error you encountered?

src/REISE.jl
dongqi-wu: > If the code is working, should we still be exiting the stdout redirection here?

Yes. I wasn't planning to put this version on PR before the meeting, so I left a lot of debug lines in. I will remove them in a later commit.

src/model.jl
dongqi-wu: > This is probably test code that can be removed?

right.

src/model.jl
dongqi-wu: > You mentioned on the call that there was some problem you encountered with the `_make_sets` function. Could you please tell us a little more about what you were trying to do, and what error you encountered?

For example, I have experimented putting line 264-268: 
>          flexible_load_bus_map = sparse(
>                sets.flexible_bus_idx_up, 1:sets.num_flexible_bus_up, 1, sets.num_bus, sets.num_flexible_bus_up
>            )::SparseMatrixCSC
in to `_make_sets()` function and in type.jl add a new field `flexible_load_bus_map`. But the code gives a strange error when running in to the end of `_make_sets()` function where the sets is created `sets = Sets(; ... )`. :
image

My workaround is putting the calculation inside fcn ` _add_constraint_power_balance!()` and everything works as expected.  This does not affect the flow and functionality of the code but I am just curious about what is causing it.

danielolsen: This message is due to the `try`/`finally` logic within **gurobi.jl**. If an error is encountered within the l.14 call to `run_scenario`, it is caught and ignored by the `try` block, then within the `finally` block, the variable `m` has not been created and therefore can't be cleaned up. If you temporarily comment out the `try`/`finally`/`end` statements in l.13, l.16, and l.20, then you will be able to see the exception and can debug.

This logic is not especially clear, and I think is an artifact of an earlier version of the Gurobi.jl package, which only required that the Gurobi environment be finalized, not the JuMP model itself.

src/types.jl
```diff
@@ -102,6 +103,11 @@ Base.@kwdef struct Sets
     # Segments
     num_segments::Int64
     segment_idx::UnitRange{Int64}
+    ## Demand Flexibility
+    flexible_bus_idx_up::Union{Array{Int64,1}, Nothing}
+    flexible_bus_idx_dn::Union{Array{Int64,1}, Nothing}
+    num_flexible_bus_up::Int64
+    num_flexible_bus_dn::Int64
```
lanesmith: Since we are assuming the buses with demand flexibility have both an up and a down component, should we have only have one `flexible_bus_idx` and one `num_flexible_bus`?

src/types.jl
```diff
@@ -52,6 +52,7 @@ Base.@kwdef struct DemandFlexibility
     enabled::Bool
     interval_balance::Bool
     rolling_balance::Bool
+    cost_input_type::String
```
lanesmith: Could this be renamed to something like `granularity`?

lanesmith:
As we mentioned yesterday, it would probably be good to include some checks on the input data in `read.jl`. Some checks that come to mind are making sure that:
- the new parameter only provides `BUS` or `AREA` as an input (also, not sure if there's a way in Julia to make sure that this string is normalized after reading the input to recognize different forms of the capitalization, i.e. inputting `BUS`, `Bus`, or `bus` are all recognized; I believe there are methods for doing this in Python) (EDIT: it appears that using `uppercase()` on the input value can act as a normalizer by capitalizing every letter in the string)
- the profiles for the demand flexibility and demand flexibility cost profiles (if included) have the same columns

src/model.jl
```diff
@@ -180,11 +186,27 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
                                * "Did you forget to linearize_gencost?")
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
+    # Demand flexibility, additional info for input->bus conversion
+    demand_flexibility_enabled = isa(demand_flexibility, DemandFlexibility) && demand_flexibility.enabled
```
lanesmith: I'm not sure this is necessary. I believe `demand_flexibility` will always be of type `DemandFlexibility` (based on the call in `REISE.jl` and the definition in `read.jl`, even if there are no demand flexibility files present), so we shouldn't need the first check.

src/model.jl
```diff
@@ -180,11 +186,27 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
...
+    if demand_flexibility_enabled && demand_flexibility.cost_input_type == "BUS"
+        flexible_bus_str_up = names(demand_flexibility.flex_amt_up)[2:end]
+        flexible_bus_str_dn = names(demand_flexibility.flex_amt_dn)[2:end]
+        flexible_bus_id_up = [parse(Int64, bus) for bus in flexible_bus_str_up]
+        flexible_bus_id_dn = [parse(Int64, bus) for bus in flexible_bus_str_dn]
+        flexible_bus_idx_up = [bus_id2idx[bus] for bus in flexible_bus_id_up]
+        flexible_bus_idx_dn = [bus_id2idx[bus] for bus in flexible_bus_id_dn]
+        num_flexible_bus_up = length(flexible_bus_idx_up)
+        num_flexible_bus_dn = length(flexible_bus_idx_dn)
```
lanesmith: We could probably have a nested `if`-`else` statement within the initially check for `demand_flexibility.enabled` that sets the demand flexibility sets for both `BUS` and `AREA`. For `BUS` the implementation would be similar to what you have now; for `AREA`, the demand flexibility sets could be set equal to those for the load buses. I believe this would remove the need for many of the `if`-`else` statements throughout `model.jl` (at least once the issue with `_make_sets` is resolved).

src/model.jl
```diff
@@ -335,27 +379,51 @@ function _add_constraints_demand_flexibility!(
...
-            sum(
-                m[:load_shift_up][i, j] - m[:load_shift_dn][i, j]
-                for j in k:(k + demand_flexibility.duration)
-            ) >= 0,
-        )
+        if demand_flexibility.cost_input_type == "AREA"
+            JuMP.@constraint(
+                m,
+                rolling_load_balance[
+                    i in 1:sets.num_load_bus,
```
lanesmith: For example, if we go with my suggestion from above, we could remove the `if`-`else` statement and just have this be `i in 1:sets.num_flexible_bus,`

src/model.jl
```diff
@@ -263,11 +299,19 @@ function _add_constraint_load_shed!(
...
+            demand_for_load_shed = JuMP.@expression(
+                m,
+                [i=1:sets.num_load_bus, j=1:interval_length],
+                demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
+            )
+        elseif demand.flexibility.cost_input_type == "BUS"
+            demand_for_load_shed = JuMP.@expression(
+                m,
+                [i=1:sets.num_load_bus, j=1:interval_length],
+                demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
```
lanesmith: Do we need a mapping here between the set of load buses and the set of demand flexibility buses? Unless the sets are equal (i.e., the `AREA` configuration), I believe `i` will exceed the index range for `load_shift_up` and `load_shift_dn` (and the buses would probably be somewhat misaligned, but exceeding the index range would likely be the thing that throws the error).

src/model.jl
```diff
@@ -91,23 +91,29 @@ function _make_bus_demand_flexibility_amount(
...
+        )
+        bus_demand_flex_amt_dn = permutedims(
+            simulation_demand_flex_amt_dn * zone_to_bus_shares
+        )
+    # input is already per-bus, copy to new container
+    elseif demand_flexibility.cost_input_type == "BUS"
+        bus_demand_flex_amt_up = transpose(simulation_demand_flex_amt_up)
+        bus_demand_flex_amt_dn = transpose(simulation_demand_flex_amt_dn)
+    else
+        println("Warning! Invalid demand flexibility parameter cost_input_type!")
```
lanesmith: If we check this in `read.jl`, I don't believe we will need this check anymore.

src/model.jl
```diff
@@ -91,23 +91,29 @@ function _make_bus_demand_flexibility_amount(
...
+        bus_demand_flex_amt_up = permutedims(
+            simulation_demand_flex_amt_up * zone_to_bus_shares
+        )
+        bus_demand_flex_amt_dn = permutedims(
+            simulation_demand_flex_amt_dn * zone_to_bus_shares
+        )
+    # input is already per-bus, copy to new container
+    elseif demand_flexibility.cost_input_type == "BUS"
+        bus_demand_flex_amt_up = transpose(simulation_demand_flex_amt_up)
+        bus_demand_flex_amt_dn = transpose(simulation_demand_flex_amt_dn)
```
lanesmith: I'm not fully clear on the merits of using `permutedims` over `transpose`, but we seem to almost exclusively use `permutedims`. I know that `permutedims` and `transpose` yield the same result but return different data types: I believe that `permutedims` returns a value that is of type `Array` while `transpose` returns a value that is of type `LinearAlgebra` (or something like that). This shouldn't be a problem because we are not inputting `bus_demand_flex_amt_up` or `bus_demand_flex_amt_dn` into any other functions, so I don't believe there should be a problem with conflicting input types. Is there a reason we generally use `permutedims`?

danielolsen: I think `transpose` returns a view into the array, while `permutedims` returns a new object, so for safety we use `permutedims` so that if we modify the result, we don't modify the upstream object.

src/types.jl
dongqi-wu: yes if that is what's desired.. One of my questions was asking if the assumption should be made or not. It looks like it is the way to go and I will remove it soon.

src/types.jl
dongqi-wu: I will change it to input_granularity

src/model.jl
dongqi-wu: understood. I will change it

src/model.jl
dongqi-wu: right this seems to be a mistake. the mapping used in _add_constraint_power_balance() should be copied here as well. hopefully it should be moved to _make_sets() so it only needs to be computed 1 time.

src/model.jl
dongqi-wu: sure I will change them to permutedims

src/model.jl
dongqi-wu: actually this seems to be required. I just recalled the reason I added this was in the end of solving optimization, the program calls make_sets() again without inputs, if I don't do isa() it will give an error.
> function _make_sets(case::Case)::Sets
>    _make_sets(case, nothing, nothing)
> end

lanesmith: I still don't think it is required to be `nothing`, but I'm glad you pointed this out as it appears to have caught something we had missed. In the code snippet that you call out, the function is used in `query.jl` to create the sets that are needed for storing the results in a comprehensible way. Before this PR, we didn't have a `demand_flexibility` input because we just used `load_bus_map` when storing the bus-level demand flexibility deviations (see [here](https://github.com/dongqi-wu/REISE.jl-dqwu/blob/4bcb153a01169e10c82ebece81ea6bc9b77c10e1/src/query.jl#L72)); a `storage` input is also not needed because we don't use any of the storage-related sets there. Since the set of flexible load buses is no longer necessarily equal to the set of load buses, we will need to pass in `flexible_load_bus_map`, which is tied to `demand_flexibility`. So, we'll need to change the snippet that you shared above to become:

```
function _make_sets(case::Case, demand_flexibility::DemandFlexibility)::Sets
    _make_sets(case, nothing, demand_flexibility)
end
```

The code in `query.jl` will also need to be updated accordingly. We will also no longer need `nothing` to be a potential input type.

BainanXia: @lanesmith According to l.147 it seems it can be `nothing` as well.

src/model.jl
BainanXia: Should we prevent user from specifying `AREA` and `BUS` with overlaps, i.e. some of the buses are within the `AREA` the demand flexibility of which is supposed to be defined in terms of `AREA`?

src/model.jl
lanesmith: @BainanXia I believe that was put there to enable the functionality in L.142-L.144. Since we need to pass `demand_flexibility` in that function, `nothing` is no longer needed as a potential input type (and should no longer need to be checked).

BainanXia: @lanesmith I'm not familiar with Julia. What is l.142 - l.144 doing? It seems to me that it just adds a wrapper of `_make_sets` function.

lanesmith: It appears to me that it is a way to create only a subset of the sets that are created in the full `_make_sets` function. This truncated version of `_make_sets` (specified in L.142-L.144) is (I believe) only used in `query.jl`, where the sets associated with `case` are used but the sets associated with `storage` are not used. This makes it so that some of the computations required to determine the storage-related sets are not performed, which should make it faster (though I'm not sure how much faster). @danielolsen could probably give a better answer as to whether or not the function specified in L.142-L.144 is doing more.

Having looked at this a little more, I wonder if we would be able to create `sets` once and take advantage of our `Sets` struct. Could it be called once in `REISE.jl`?

src/model.jl
lanesmith: I'm not sure I understand your question. The user only specifies `AREA` and `BUS` once at input (via `demand_flexibility_parameters.csv`). This helps determine the granularity of the demand flexibility data (e.g., if the user specifies `BUS`, all demand flexibility data that is input will be bus-level). Is that what you were asking?

BainanXia: I see. So we don't allow the user to have both area and bus simultaneously, right? Hence, in certain situation, assuming in a use case, we would like to have all buses in CA plus some buses in WA with demand flexibility defined, the user has to choose `BUS` then put down all buses in CA and the subset of buses in WA in columns.

src/model.jl
BainanXia: But both functions are defined in the same domain with same name, will the full `_make_sets` overrides the truncated one?

src/model.jl
lanesmith: Correct. I think the functionality you describe sounds great, but would that be better left for a future PR?

BainanXia: @lanesmith Does it imply a major refactor of the code in this PR? If so, probably we should do it here.

src/model.jl
lanesmith: This is where we approach the end of my limited Julia knowledge haha. Doing a simple test (and based on the functions working in our simulations), it appears that Julia can tell the difference between tests of the same name so long as they have different inputs. I'm not certain that we can do this in Julia, but if we were to do this in a more Pythonic way, we could probably set `storage=nothing` as a default in the main `_make_sets` function.

src/model.jl
lanesmith: I don't think it would be a major refactor, it would just be a bigger change in the way that the granularity is input and we would need a lot more checks in `read_demand_flexibility` so that we prevent the overlap problem that you described above.

BainanXia: Cool, then let's save it to a separate PR.

src/model.jl
```diff
@@ -299,19 +291,12 @@ function _add_constraint_load_shed!(
...
-                m,
-                [i=1:sets.num_load_bus, j=1:interval_length],
-                demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
-            )
-        end
+        demand_for_load_shed = JuMP.@expression(
+            m,
+            [i in sets.flexible_bus_idx, j=1:interval_length],
+            demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
+        )
```
lanesmith: EDIT: Reposting this review below as it didn't recognize that I was reviewing the new code.

I think we might need to rethink how we update this expression. I believe this expression overwrites the previous `demand_for_load_shed` expression, which is a problem when the set of flexible load buses is not equal to the set of load buses. When the set of flexible load buses is not equal to the set of load buses, this second iteration of `demand_for_load_shed` makes it so that there is only a value for buses that are included in the set of flexible load buses. This is inaccurate because `load_shed` occurs at each of the load buses (`load_shed` is not demand flexibility here, but rather a means to allow the simulation to avoid infeasibilities).

One solution that I can think of is doing something like we do in `_add_constraint_power_balance!`, where we could multiply the `load_shift_up` and `load_shift_dn` variables by some map between the load buses and the flexible load buses. Maybe something like this:

```
demand_for_load_shed = JuMP.@expression(
    m,
    demand_for_load_shed 
    + {map between load buses and flexible load buses} * m[:load_shift_up] 
    - {map between load buses and flexible load buses} * m[:load_shift_dn],
)
```

where `{map between load buses and flexible load buses}` would probably be created in a similar fashion to `flexible_load_bus_map` and may not need to be a part of `sets` (if this is the only place it is used).

src/model.jl
```diff
@@ -265,9 +293,10 @@ function _add_constraint_load_shed!(
     if demand_flexibility.enabled
         demand_for_load_shed = JuMP.@expression(
             m,
-            [i=1:sets.num_load_bus, j=1:interval_length],
+            [i in sets.flexible_bus_idx, j=1:interval_length],
             demand_for_load_shed[i, j] + m[:load_shift_up][i, j] - m[:load_shift_dn][i, j],
         )
```
lanesmith: I think we might need to rethink how we update this expression. I believe this expression overwrites the previous `demand_for_load_shed` expression, which is a problem when the set of flexible load buses is not equal to the set of load buses. When the set of flexible load buses is not equal to the set of load buses, this second iteration of `demand_for_load_shed` makes it so that there is only a value for buses that are included in the set of flexible load buses. This is inaccurate because `load_shed` occurs at each of the load buses (`load_shed` is not demand flexibility here, but rather a means to allow the simulation to avoid infeasibilities).

One solution that I can think of is doing something like we do in `_add_constraint_power_balance!`, where we could multiply the `load_shift_up` and `load_shift_dn` variables by some map between the load buses and the flexible load buses. Maybe something like this:

```
demand_for_load_shed = JuMP.@expression(
    m,
    demand_for_load_shed 
    + {map between load buses and flexible load buses} * m[:load_shift_up] 
    - {map between load buses and flexible load buses} * m[:load_shift_dn],
)
```

where `{map between load buses and flexible load buses}` would probably be created in a similar fashion to `flexible_load_bus_map` and may not need to be a part of `sets` (if this is the only place it is used).

src/model.jl
```diff
@@ -291,12 +291,17 @@ function _add_constraint_load_shed!(
         bus_demand[sets.load_bus_idx[i], j],
     )
     if demand_flexibility.enabled
+        flexible_load_bus_idx = indexin(sets.flexible_bus_idx, sets.load_bus_idx)
+        flexible_load_map = sparse(
+                flexible_load_bus_idx, 1:num_flexible_bus, 1, sets.num_load_bus, num_flexible_bus
```
lanesmith: Should `num_flexible_bus` be `sets.num_flexible_bus`?

src/model.jl
```diff
@@ -291,12 +291,17 @@ function _add_constraint_load_shed!(
...
+            )::SparseMatrixCSC
+        #
```
lanesmith: Were you intending to leave a comment here? If not, this comment can probably be removed.

src/model.jl
```diff
@@ -232,11 +260,11 @@ function _add_constraint_power_balance!(
...
     if demand_flexibility.enabled
         injections = JuMP.@expression(
-            m, injections + sets.load_bus_map * m[:load_shift_dn]
+            m, injections + sets.flexible_load_bus_map * m[:load_shift_dn]
         )
         withdrawals = JuMP.@expression(
-            m, withdrawals + sets.load_bus_map * m[:load_shift_up]
-        )
+            m, withdrawals + sets.flexible_load_bus_map * m[:load_shift_up]
+        )
```
danielolsen: Unnecessary whitespace change.

src/model.jl
```diff
@@ -338,24 +371,25 @@ function _add_constraints_demand_flexibility!(
...
-                i in 1:sets.num_load_bus,
+                i in 1:sets.num_flexible_bus,
                 k in 1:(interval_length - demand_flexibility.duration)
             ],
             sum(
                 m[:load_shift_up][i, j] - m[:load_shift_dn][i, j]
                 for j in k:(k + demand_flexibility.duration)
             ) >= 0,
-        )
+        )
```
danielolsen: Unnecessary whitespace change.

src/model.jl
```diff
@@ -338,24 +371,25 @@ function _add_constraints_demand_flexibility!(
...
+                i in 1:sets.num_flexible_bus,
                 k in 1:(interval_length - demand_flexibility.duration)
             ],
             sum(
                 m[:load_shift_up][i, j] - m[:load_shift_dn][i, j]
                 for j in k:(k + demand_flexibility.duration)
             ) >= 0,
-        )
+        )            
+        #end
```
danielolsen: Old comment?

src/model.jl
```diff
@@ -338,24 +371,25 @@ function _add_constraints_demand_flexibility!(
...
         println("interval load balance: ", Dates.now())
         JuMP.@constraint(
             m,
-            interval_load_balance[i in 1:sets.num_load_bus],
+            interval_load_balance[i in 1:sets.num_flexible_bus],
             sum(
                 m[:load_shift_up][i, j] - m[:load_shift_dn][i, j] for j in 1:interval_length
             ) >= 0,
-        )
+        )
```
danielolsen: Unnecessary whitespace change.

src/model.jl
```diff
@@ -134,11 +140,11 @@ end
 
 
 function _make_sets(case::Case)::Sets
-    _make_sets(case, nothing)
+    _make_sets(case, nothing, nothing)
 end
 
 
-function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
+function _make_sets(case::Case, storage::Union{Storage,Nothing}, demand_flexibility::Union{DemandFlexibility, Nothing})::Sets
```
danielolsen: I wonder if we could combine these:
```julia
_make_sets(
    case::Case,
    storage::Union{Storage,Nothing}=nothing,
    demand_flexibility::Union{DemandFlexibility, Nothing}=nothing,
)::Sets
```

src/model.jl
```diff
@@ -669,16 +700,15 @@ function _build_model(
...
         )
-
-        # The amount of demand that is added to the base load
+        # The amount of demand that is added from the base load
         JuMP.@variable(
-            m, 
-            0 <= load_shift_up[i in 1:sets.num_load_bus, j in 1:interval_length]
-                <= bus_demand_flex_amt_up[sets.load_bus_idx[i], j],
+            m,
+            0 <= load_shift_up[i in 1:sets.num_flexible_bus, j in 1:interval_length]
```
danielolsen: Unnecessary whitespace at end of line.

src/model.jl
lanesmith: In a similar vein, it might be nice to call `_make_sets` in `REISE.jl` (within the `run_scenario` function) so that the sets only need to be created once. Right now, we create the sets in `loop.jl`, `model.jl`, and `query.jl`. I don't believe the information in the sets changes as the simulation progresses.

src/model.jl
```diff
@@ -180,11 +186,31 @@ function _make_sets(case::Case, storage::Union{Storage,Nothing})::Sets
                                * "Did you forget to linearize_gencost?")
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
+    # Demand flexibility, additional info for input->bus conversion
+    if demand_flexibility.enabled
+        if demand_flexibility.input_granularity == "BUS"
+            # assume each flexible bus can go up/dn, so only use the Dataframe for up
+            flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
```
dmuldrew, line 191: Might want to include a quick comment for why the index starts at 2 here, I'm guessing the CSV is expected to have a line with column names?

danielolsen: This follows from my previous (probably bad) convention. We could probably make this cleaner throughout the whole function by indicating that the first column is an index when we first load the CSVs.

src/model.jl
dongqi-wu: @danielolsen I should have fixed the white space problem in the last commit. I am using jupyter to run julia codes and the editor does not visually distinguish white spaces from the background so I didn't notice, sorry for that. Also I apologize for addressing these comments so late, I spent all the time in the past 10 days on working on a TSG special issue submission of which the deadline just passed..

danielolsen: Good luck on the submission! PR #156 will introduce automated style checking, so this should be more obvious in the future.

src/model.jl
```diff
@@ -263,10 +289,15 @@ function _add_constraint_load_shed!(
         bus_demand[sets.load_bus_idx[i], j],
     )
     if demand_flexibility.enabled
+        flexible_load_bus_idx = indexin(sets.flexible_bus_idx, sets.load_bus_idx)
+        flexible_load_map = sparse(
+                flexible_load_bus_idx, 1:sets.num_flexible_bus, 1, sets.num_load_bus, sets.num_flexible_bus
+            )::SparseMatrixCSC
```
danielolsen, line 295: Is this duplicated from the calculation in `_make_sets`? Within the expression definition below, can we use `sets.flexible_load_bus_map` instead?

danielolsen (approved):
The logic makes sense, and the code runs successfully using the test inputs provided. There's one place where it seems that we're re-calculating something that we already have; once that is removed, or if in fact we're actually calculating something different, I think this is ready to merge.

BainanXia (approved):

src/model.jl
dongqi-wu: These are not the same. The code shown here uses the index of flexible buses in load buses, while the _make_set() fcn computes the index of flexible buses in all buses. The load bus vector is a subset of the bus vector so the indices would be different.

danielolsen: Ah, good catch.

lanesmith (approved):
It looks like all of my comments have been addressed with the exception of including checks on the profile inputs. That functionality is probably better left to #157 though.

PR #156, opened 2021-09-22
ci: format Julia code to Blue style, add CI to check new commits

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Closes #152.

### What the code is doing
- Existing code is reformatted (should be no change to functionality).
- A `.git-blame-ignore-revs` file is added to ignore the reformat commit when using git blame.
- `lint.yml` is modified to add an automated check of Julia formatting.
- `.JuliaFormatter.toml` specifies that the automated check is against the Blue style (unlike black, which is the reformatter and the style, `JuliaFormatter` is a reformatter which can use one of many styles, so we have to tell it that we want `Blue`).

### Testing
Tested via pushes to this branch.

### Usage Example
```julia
using JuliaFormatter
format(".")
```

### Time estimate
15 minutes.

PR #159, opened 2021-10-08
fix: _make_sets invocation

jenhagg:
### Purpose
Attempt to fix an error we get launching a simulation. This is the (truncated) output from `scenario.check_progress()` 
```
  'Julia exception: MethodError: no method matching _make_sets(::REISE.Case, ::REISE.Storage, ::REISE.DemandFlexibility)',
  'Closest candidates are:',
  '_make_sets(::REISE.Case; storage, demand_flexibility) at /app/src/model.jl:133',
  'Stacktrace:',
  '[1] interval_loop(::Type{T} where T, ::Dict{String,Any}, ::Dict{Any,Any}, ::Int64, ::Int64, ::Int64, ::String, ::String) at /app/src/loop.jl:47',
  '[2] (::REISE.var"#116#117"{Int64,Int64,Int64,String,DataType,Dict{String,Any}})() at /app/src/REISE.jl:84',

```

### What the code is doing
First I tried changing the signature back to a comma, but the formatter doesn't like that.  So instead we can update usages to keyword arguments.

### Testing
Launched a scenario in a container.

### Time estimate
2 min

### DIFF
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -44,7 +44,7 @@ function interval_loop(
     case = model_kwargs["case"]
     storage = model_kwargs["storage"]
     demand_flexibility = model_kwargs["demand_flexibility"]
-    sets = _make_sets(case, storage, demand_flexibility)
+    sets = _make_sets(case; storage=storage, demand_flexibility=demand_flexibility)
     unused_load_shed_intervals_turnoff = 14
     # Start looping
     for i in 1:n_interval
--- a/src/model.jl
+++ b/src/model.jl
@@ -643,7 +643,7 @@ function _build_model(
     hour_idx = 1:interval_length
     end_index = start_index + interval_length - 1
     # Sets - static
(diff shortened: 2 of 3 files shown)


BainanXia (approved):
LGTM. This will cause an issue when launching a scenario as well or only matters when call `scenario.check_progress()`?

danielolsen (approved):
I think I see what happened here. When we refactored the `_make_sets` function within #155, we added inputs with default values, and then in #156 the styler said 'since there are default values, these parameters need to be passed as keyword arguments' (which is what the `;` signifies), but didn't change the invocations in the existing code.

Thanks for the quick fix.

jenhagg:
> LGTM. This will cause an issue when launching a scenario as well or only matters when call `scenario.check_progress()`?

It happened while launching a scenario. The check progress just gets the stdout from that call since we run a new process in the container.

PR #160, opened 2021-10-22
feat: support for mixed flexibility input granularity (#157)

### LINKED ISSUE #157: Feature request: Support for Mixed Input Granularity
# :rocket:

- [ ] Is your feature request essential for your project?

In the newest version in #155 the user specified input files `demand_flexibility_up/dn.csv` and `demand_flexibility_cost_up/dn.csv` the user has two options (from a field in demand_flexibility_parameters.csv): `AREA` or `BUS`. The column headers in the input csvs must be corresponding to the parameter. (e.g. if the parameter is `AREA` then the user must give a column of flexibility/cost values for every load zones in the network).

I plan to add support for letting these two types of columns to co-exist. The user can choose to give flexibility/cost data for any number of load zones and any number of load buses in the same csv file. This will allow users to specify some rough numbers for large part of the system and high resolution numbers for a few buses to be studied.

Write additional function and codes in REISE.jl immediately after reading the demand_flexibility struct to convert the input dataframes to the proper format in which all columns correspond to load buses.

e. g.:
Name: preprocess_flexibility
Inputs: demand_flexibility
Description: Go through `demand_flexibility.flex_amt`, `demand_flexibility.cost_up` and `demand_flexibility.cost_dn` which are read from raw input files, and replace them with dataframe that has all bus-level data with shape (hour_num, flexible_bus_num)

Since this approach is more general and can support both type of input files in the current version, the parameter 'input_granularity' will be deleted and the code related to area-level input and distinguishing between `BUS` and `AREA` will be deleted.

### Describe alternatives you've considered, if relevant
Two concerns might be encountered in this implementation

1) When bus IDs and load zone IDs are not unique. For example if there exist a bus called `123` and also a load zone called `123` they cannot be distinguished from the csv header. One way to get around this is to make every ID unique, and another way is to add prefix in the csv headers such as `bus.123` and `area.123`

2) When any load bus get specified two times, one in the area they belong and another with their bus number. For example, if load `123` belong to area `A`, and the input csv has columns for both `123` and `A`. The simplest way is to throw an error, while alternatively the final flexibility amount for bus `123` can be interpreted as the sum of values in column `123` and its part from column `A`. Another way might be re-distributing flexibility values in column `A` to other buses in `A` excluding bus `123`, to make sure the total flexibility in area `A` still sum up to the values in that column.

dongqi-wu:

### Purpose
Implement changed proposed in #157 

### What the code is doing
inspect the raw input dataframe for demand flexibility and its cost, convert zone-level flexibility to bus-level and concatenate with the bus-level flexibility input columns 

### Where to look
the core code is added to bottom read.jl

### Time estimate
15-30 min

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -70,10 +70,11 @@ function _make_bus_demand_flexibility_amount(
     simulation_demand_flex_amt_dn = Matrix(
         demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
     )
-    # input is already per-bus, copy to new container
-    bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
-    bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
-    return (bus_demand_flex_amt_up, bus_demand_flex_amt_dn)
+
+    return (
+        permutedims(simulation_demand_flex_amt_up),
+        permutedims(simulation_demand_flex_amt_dn),
+    )
 end
 
 """
--- a/src/model.jl
+++ b/src/model.jl
@@ -37,28 +37,6 @@ function _make_branch_map(case::Case)::SparseMatrixCSC
 end
 
 """
-    _make_bus_demand_weighting(case)
-
-Given a Case object, build a sparse matrix that indicates the weighting of each bus in 
-    each zone.
-"""
-function _make_bus_demand_weighting(case::Case)::SparseMatrixCSC
-    bus_idx = 1:length(case.busid)
-    bus_df = DataFrames.DataFrame(;
-        name=case.busid, load=case.bus_demand, zone=case.bus_zone
-    )
-    zone_demand = DataFrames.combine(DataFrames.groupby(bus_df, :zone), :load => sum)
-    zone_list = sort(collect(Set(case.bus_zone)))
-    zone_idx = 1:length(zone_list)
-    zone_id2idx = Dict(zone_list .=> zone_idx)
-    bus_df_with_zone_load = DataFrames.innerjoin(bus_df, zone_demand; on=:zone)
-    bus_share = bus_df[:, :load] ./ bus_df_with_zone_load[:, :load_sum]
-    bus_zone_idx = Int64[zone_id2idx[z] for z in case.bus_zone]
-    zone_to_bus_shares = sparse(bus_zone_idx, bus_idx, bus_share)::SparseMatrixCSC
-    return zone_to_bus_shares
-end
-
-"""
     _make_bus_demand(case)
 
 Given a Case object, build a matrix of demand by (bus, hour) for this interval.
@@ -92,21 +70,9 @@ function _make_bus_demand_flexibility_amount(
     simulation_demand_flex_amt_dn = Matrix(
         demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
     )
-    # convert per-area input to per-bus values
-    if demand_flexibility.input_granularity == "AREA"
-        bus_demand_flex_amt_up = permutedims(
-            simulation_demand_flex_amt_up * zone_to_bus_shares
-        )
-        bus_demand_flex_amt_dn = permutedims(
-            simulation_demand_flex_amt_dn * zone_to_bus_shares
-        )
-        # input is already per-bus, copy to new container
-    elseif demand_flexibility.input_granularity == "BUS"
-        bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
-        bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
-    else
-        println("Warning! Invalid demand flexibility parameter input_granularity!")
-    end
+    # input is already per-bus, copy to new container
+    bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
+    bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
     return (bus_demand_flex_amt_up, bus_demand_flex_amt_dn)
 end
 
@@ -179,22 +145,18 @@ function _make_sets(
     )
     num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
     segment_idx = 1:num_segments
-    # Demand flexibility, additional info for input->bus conversion
-    if demand_flexibility.enabled
-        if demand_flexibility.input_granularity == "BUS"
-            # assume each flexible bus can go up/dn, so only use the Dataframe for up
-            flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
-            flexible_bus_id = [parse(Int64, bus) for bus in flexible_bus_str]
-            flexible_bus_idx = [bus_id2idx[bus] for bus in flexible_bus_id]
-            num_flexible_bus = length(flexible_bus_idx)
-            flexible_load_bus_map = sparse(
-                flexible_bus_idx, 1:num_flexible_bus, 1, num_bus, num_flexible_bus
-            )::SparseMatrixCSC
-        elseif demand_flexibility.input_granularity == "AREA"
-            flexible_bus_idx = load_bus_idx
-            num_flexible_bus = num_load_bus
-            flexible_load_bus_map = load_bus_map
-        end
+    # Demand flexibility, additional info for bus <--> load <--> flexible load conversion
+    demand_flexibility_enabled =
+        isa(demand_flexibility, DemandFlexibility) && demand_flexibility.enabled
+    if demand_flexibility_enabled
+        # assume each flexible bus can go up/dn, so only use the Dataframe for up
+        flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
+        flexible_bus_id = [parse(Int64, bus) for bus in flexible_bus_str]
+        flexible_bus_idx = [bus_id2idx[bus] for bus in flexible_bus_id]
+        num_flexible_bus = length(flexible_bus_idx)
+        flexible_load_bus_map = sparse(
+            flexible_bus_idx, 1:num_flexible_bus, 1, num_bus, num_flexible_bus
+        )::SparseMatrixCSC
     else
         flexible_bus_idx = nothing
         num_flexible_bus = 0
@@ -379,7 +341,7 @@ function _add_constraints_demand_flexibility!(
         println("rolling load balance, first window: ", Dates.now())
         JuMP.@constraint(
             m,
-            rolling_load_balance_first[i in 1:(sets.num_load_bus)],
+            rolling_load_balance_first[i in 1:(sets.num_flexible_bus)],
             sum(
                 m[:load_shift_up][i, j] - m[:load_shift_dn][i, j] for
                 j in 1:(demand_flexibility.duration - 1)
--- a/src/read.jl
+++ b/src/read.jl
@@ -298,7 +298,7 @@ function reformat_demand_flexibility_input(
             ErrorException(
                 "The flexible bus/load zone specified in the up/down " *
                 "input csvs do not match. Please check the input files to make sure " *
-                "every flexible bus orload zone has corresponding columns in both " *
+                "every flexible bus or load zone has corresponding columns in both " *
                 "flexibility csvs.",
             ),
         )
--- a/src/read.jl
+++ b/src/read.jl
@@ -318,6 +318,11 @@ function reformat_demand_flexibility_input(
         # dataframe in the corresponding field
         fh = getfield(demand_flexibility, Symbol(field))
 
+        # skip if no cost
+        if isnothing(fh)
+            continue
+        end
+
         # headers
         flexible_str = names(fh)[2:end]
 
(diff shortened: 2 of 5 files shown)


src/read.jl
```diff
@@ -261,3 +255,99 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+    bus_idx = 1:length(case.busid)
+    bus_df = DataFrames.DataFrame(
+        name=case.busid, load=case.bus_demand, zone=case.bus_zone
+    )
+    zone_demand = DataFrames.combine(
+        DataFrames.groupby(bus_df, :zone), :load => sum
+    )
+    zone_list = sort(collect(Set(case.bus_zone)))
+    zone_idx = 1:length(zone_list)
+    zone_id2idx = Dict(zone_list .=> zone_idx)
```
BainanXia, line 274: I presume the reason we would like to transform `id` to index is Julia can't do element-wise matrix/dataframe multiplications based on column names, instead the order of rows/columns must be aligned?

dongqi-wu: I think here the matrix to be indexed is not dataframe. (a matrix does not have header/index row/col to be used with names).. BTW even for dataframes (see the fcn below) I used the same approach (id to idx) because I feel more comfortable with that. If its more "standard" to index be names when available I can also fix that.

src/read.jl
```diff
@@ -261,3 +255,99 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+"""
+    _make_bus_demand_weighting(case)
+
+Given a Case object, build a sparse matrix that indicates the weighting of each bus in 
+    each zone.
+"""
+function _make_bus_demand_weighting(case::Case)::SparseMatrixCSC
+    bus_idx = 1:length(case.busid)
+    bus_df = DataFrames.DataFrame(
+        name=case.busid, load=case.bus_demand, zone=case.bus_zone
```
BainanXia, line 269: `name=case.busid` is never used here, right?

src/read.jl
```diff
@@ -261,3 +255,99 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+    )
+    zone_demand = DataFrames.combine(
+        DataFrames.groupby(bus_df, :zone), :load => sum
+    )
+    zone_list = sort(collect(Set(case.bus_zone)))
+    zone_idx = 1:length(zone_list)
+    zone_id2idx = Dict(zone_list .=> zone_idx)
+    bus_df_with_zone_load = DataFrames.innerjoin(bus_df, zone_demand, on=:zone)
+    bus_share = bus_df[:, :load] ./ bus_df_with_zone_load[:, :load_sum]
```
BainanXia: Not sure how Julia manages memory usage. It seems we don't need to create new data frames, instead all the calculations can be done on the fly: 
- start with a data frame with bus demand column and zone id -> 
- group rows by zone id then summation by bus demand (this potentially gives us a dictionary like object) -> 
- add another column with the values being demand for each row divided by the corresponding zonal demand from the dict object
Maybe I'm wrong due to Pythonic thinking. Just bring this up into discussion.

src/types.jl
```diff
@@ -41,7 +41,7 @@ Base.@kwdef struct Storage
     sd_table::DataFrames.DataFrame
 end
 
-Base.@kwdef struct DemandFlexibility
+Base.@kwdef mutable struct DemandFlexibility
```
BainanXia: I presume we need it to be `mutable` because we transform all columns into buses in place, right?

src/read.jl
```diff
@@ -261,3 +255,99 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                )
+        end
+        
+        # pick out zone columns in raw input
+        zone_input_cols = DataFrames.select(fh, [i+1 for i in zone_columns_idx])
+        # list index of each zone in input file in the sorted list of zones
+        zone_cols_idx = findall(x -> x in flexible_zone_id, zone_list)
+
+        # convert zone-level to bus-level
+        converted_zone_flexibility = Matrix(zone_input_cols) * zone_to_bus_shares[zone_cols_idx,:]
```
BainanXia: Again, we transform ids into sorted indices for both buses and zones to align them before multiplication. Just curious whether there is a way in Julia to conduct matrix multiplications by column names (ids) rather than order (indices)?

BainanXia:
I've gone through the code and the logic is clear. Several high-level concerns:
- When we have same bus referred in both bus columns and zone column, instead of summation, we might want to go with the other option to keep the totals of the corresponding zone unchanged.
- Any tests are run to validate the current implementation?
- Minor issue: the linting fails. We added this workflow recently in this [PR](https://github.com/Breakthrough-Energy/REISE.jl/pull/156). Detailed instructions of running the linter locally can be found there as well.

src/read.jl
dongqi-wu: this is from the old code that was here so I kept it as it. not sure if is needed elsewhere but for my code it is not.

src/read.jl
dongqi-wu: I am not familiar with Julia.. I think @danielolsen knows more about it. I was trying to replicate what I would do with Pandas, it might not be the best approach.

src/types.jl
dongqi-wu: In the big read function the dfs are stored in flex_amt_up/dn and cost_up/dn. These values are later replaced in the new fcn so it has to be mutable. Another approach could be putting the new code inside the big read fun (i.e. construct the bus-level DF before assigning to the fields in the df struct). I thought about this but it requires parsing the case in REISE.jl before reading all input csvs which requires a change in the workflow. A better method (if exist) could be making just the fields to be mutable, but I google-d for a while and didn't find a good solution.

src/read.jl
dongqi-wu: This is also copied from the old code. Maybe we can delete the old DF after creating the new one ?

dongqi-wu:
> I've gone through the code and the logic is clear. Several high-level concerns:
> 
> * When we have same bus referred in both bus columns and zone column, instead of summation, we might want to go with the other option to keep the totals of the corresponding zone unchanged.
> * Any tests are run to validate the current implementation?
> * Minor issue: the linting fails. We added this workflow recently in this [PR](https://github.com/Breakthrough-Energy/REISE.jl/pull/156). Detailed instructions of running the linter locally can be found there as well.

1. I will change that in a later commit
2. Yes, I have created some mixed header csv, I will upload them after item 1.
3. what does the Lint do? I tried to look at the report from the formatting test and it does not show any info -- Is there any specific format standard that it tries to look for?

lanesmith:
> What does the Lint do? I tried to look at the report from the formatting test and it does not show any info -- Is there any specific format standard that it tries to look for?

The linter that failed checks the Julia code to make sure it complies with the format style. The code style that is being used is [Blue](https://github.com/invenia/BlueStyle). I believe some of the comments in #156 address how to use `JuliaFormatter` to check and format your code locally.

dongqi-wu:
image
image
I have fixed how overlapping columns are interpreted, but still get the lint error. It seems lint can't run on my pc after following #156. Is there any other package I will need to install to support the blue style checking?

src/model.jl
```diff
@@ -92,21 +70,9 @@ function _make_bus_demand_flexibility_amount(
...
-        # input is already per-bus, copy to new container
-    elseif demand_flexibility.input_granularity == "BUS"
-        bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
-        bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
-    else
-        println("Warning! Invalid demand flexibility parameter input_granularity!")
-    end
+    # input is already per-bus, copy to new container
+    bus_demand_flex_amt_up = permutedims(simulation_demand_flex_amt_up)
+    bus_demand_flex_amt_dn = permutedims(simulation_demand_flex_amt_dn)
```
lanesmith: Is it necessary to create these new variables to just compute the transposes? Could `permutedims` be applied to `simulation_demand_flex_amt_up` and `simulation_demand_flex_amt_dn` above and then instead return those two variables?

src/read.jl
```diff
@@ -261,3 +255,197 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+            # numeric ID corresponding to bus columns
+            flexible_bus_id = [parse(Int64, flexible_str[i]) for i in bus_columns_idx]
+            flexible_bus_num = length(flexible_bus_id)
+
+            # check if zone numbers are correct
+            if !all([issubset(i, zone_list) for i in flexible_zone_id])
+                @error("Invalid load zone numeric ID(s) in demand flexibility input files!")
+            elseif !all([issubset(i, case.busid) for i in flexible_bus_id])
+                @error("Invalid load bus numeric ID(s) in demand flexibility input files!")
+            end
```
lanesmith: Are we actually trying to throw an error here? If so, I'm not sure that `@error` will do that. If I remember correctly, `@error` seems to act like a fancy print statement. I believe we would need something like `throw(ErrorException("some message"))` if we actually want the simulation to stop. If we just want to notify the user of the error, then `@error` should be fine.

src/read.jl
```diff
@@ -261,3 +255,197 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+
+            # list index of each zone in input file in the sorted list of zones
+            zone_cols_idx = findall(x -> x in flexible_zone_id, zone_list)
```
lanesmith: This is related to my comment on `@error`. If the intent of `@error` is to stop the simulation, is there a functional difference between `zone_cols_idx` and `zone_columns_idx`? In other words, if the above check stops the simulation if there are invalid zones, then shouldn't `flexible_zone_id` (which is based on `zone_columns_idx`) and `zone_list` be the same?

src/read.jl
```diff
@@ -261,3 +255,197 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                    Matrix(fh[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_shares[zone_cols_idx, :]
+
+                # add bus-level columns on top of converted bus-level flexibility matrix
+                flex_bus_idx = zeros(Int64, 0)
+                for i in sets.bus_idx
+                    bus = case.busid[i]
+                    # add specified flexibility to the corresponding bus
+                    if bus in flexible_bus_id
+                        converted_zone_flexibility[:, i] += fh[string(bus)]
+                    end
```
lanesmith, line 438: I want to check my understanding of the bus-level demand flexibility distribution. This seems that the total zone flexibility is reduced according to the individual bus-level demand flexibility that is specified (which appears to track with some of the discussion from #157). However, it seems that the modified zone flexibility is then distributed to all demand buses in that zone, even the ones with otherwise specified demand flexibility. Then, the specified bus-level demand flexibility is added to the appropriate bus, which already has a portion of demand flexibility according to the `zone_to_bus_shares`. Is this accurate? If so, is that the intent?

src/read.jl
```diff
@@ -261,3 +255,197 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+
+                    # identify flexible bus by their total flexibility and append to list
+                    if any(x -> x > 0, converted_zone_flexibility[:, i])
+                        append!(flex_bus_idx, i)
+                    end
+                end
+                # for cost, the zone numbers apply to all buses except those with dedicated columns
+            else
+                # convert zone-level cost using incidence matrix
+                converted_zone_flexibility =
```
lanesmith: Should this be `converted_zone_cost`, along with the other instances of `converted_zone_flexibility` in the cost-related section?

src/read.jl
```diff
@@ -261,3 +255,197 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                end
+            end
+
+            # add header and datetime index column
+            eq_bus_df = DataFrames.DataFrame(converted_zone_flexibility[:, flex_bus_idx])
+            DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flex_bus_idx]))
+            DataFrames.insertcols!(eq_bus_df, 1, :"UTC Time" => fh["UTC Time"])
+
+            # replace the raw input df
+            #setfield!(demand_flexibility_updated, Symbol(field), eq_bus_df)
```
lanesmith: Is this old code that is no longer needed?

lanesmith:
Are there plans to implement a check to make sure that each bus has both up and down flexibility? With users having the ability to specify bus-level flexibility individually, I can see a bus easily being left out in one direction. This would create a problem in the simulation (though not necessarily a problem that would be readily obvious) since this demand flexibility model focuses on shiftable demand (i.e., failing to be flexible in one direction means the bus would not be flexible at all and, as of now, an error wouldn't be thrown).

src/model.jl
dongqi-wu: I guess in the previous version there was an if to decide how these are computed (differently based on input params). now they can be changed as you suggest.

src/read.jl
dongqi-wu: I will look into it. sorry I am not that familiar with julia

src/read.jl
dongqi-wu: this is the index in the zone list (which will be used later to index the zone-bus mapping matrix). e.g. if there are total 3 zones [1, 2, 3]; 3 zone columns [2, 3, 1] then this index would be [1, 2, 0]

src/read.jl
dongqi-wu: ah yeah I copied the code and forgot to rename it. functionally it should be the same but it can be confusing. I will rename it.

src/read.jl
dongqi-wu: I see. this seems to be a bug. The zone-to-bus-share should be computed/modified to consider the bus columns.

dongqi-wu:
> Are there plans to implement a check to make sure that each bus has both up and down flexibility? With users having the ability to specify bus-level flexibility individually, I can see a bus easily being left out in one direction. This would create a problem in the simulation (though not necessarily a problem that would be readily obvious) since this demand flexibility model focuses on shiftable demand (i.e., failing to be flexible in one direction means the bus would not be flexible at all and, as of now, an error wouldn't be thrown).

I think there was a previous conversation about this and the result is to assume every flexible bus can go either way and will be present in both input files. For one-way flexibility the user will need to explicitly set the corresponding column in one of the csvs to all 0s.

src/read.jl
lanesmith: No worries. Intuitively, you would think it would throw an error. [Here](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/read.jl#L233) is how I used `throw` in conjunction with `@error`. `@error` throws the nice-looking error message and `throw` stops the simulation and produces a stacktrace.

src/read.jl
lanesmith: Yeah, that's what I was thinking regarding the `zone_to_bus_shares`. Thanks for clarifying!

lanesmith:
> I think there was a previous conversation about this and the result is to assume every flexible bus can go either way and will be present in both input files. For one-way flexibility the user will need to explicitly set the corresponding column in one of the csvs to all 0s.

Yeah, I remember that being the assumption that we make. Since we are using shiftable demand, I think that assumption will be necessary until we implement sheddable demand (or have very specific technology demand models, which we probably don't need). Despite this assumption, my concern is that the potential for a bunch of inputs that are a mix of zones and buses presents an opportunity for user error when creating the `demand_flexibility_up` and `demand_flexibility_dn` profiles. The way the model works, there is no way that a user would know that they made an error with creating those profiles until after the simulation finished and they really dug into the bus-level demand flexibility data.

dongqi-wu:
> > I think there was a previous conversation about this and the result is to assume every flexible bus can go either way and will be present in both input files. For one-way flexibility the user will need to explicitly set the corresponding column in one of the csvs to all 0s.
> 
> Yeah, I remember that being the assumption that we make. Since we are using shiftable demand, I think that assumption will be necessary until we implement sheddable demand (or have very specific technology demand models, which we probably don't need). Despite this assumption, my concern is that the potential for a bunch of inputs that are a mix of zones and buses presents an opportunity for user error when creating the `demand_flexibility_up` and `demand_flexibility_dn` profiles. The way the model works, there is no way that a user would know that they made an error with creating those profiles until after the simulation finished and they really dug into the bus-level demand flexibility data.

so you suggest there should be a warning for that ?

lanesmith:
> so you suggest there should be a warning for that ?

My inclination would be to throw an error. If there is a mismatch between the zones and buses provided in `demand_flexibility_up` and `demand_flexibility_dn`, there is likely (a) a mistake or (b) an implementation that will not work as the user intended.

src/read.jl
lanesmith: Hmm when I test this they seem to produce the same array, even when the zone columns are not ordered. For an example with four zones (`["zone.301", "zone.302", "zone.304", "zone.303"]`), I get that both `zone_columns_idx` and `zone_cols_idx` are equal to `[1, 2, 3, 4]`.

dongqi-wu: Just confirmed the problem.. It seems this findall iteration is not keeping the original order of elements in flexible_zone_id. I had to use `zone_cols_idx = [findfirst(y -> y == x , zone_list) for x in flexible_zone_id]` to avoid losing the order. This behavior of search related fcns in Julia is so strange..

dongqi-wu:
> > so you suggest there should be a warning for that ?
> 
> My inclination would be to throw an error. If there is a mismatch between the zones and buses provided in `demand_flexibility_up` and `demand_flexibility_dn`, there is likely (a) a mistake or (b) an implementation that will not work as the user intended.

Hi Lane, I have pushed a new commit that should fix the issues in your comments. One thing I noticed is that the throw error thing is masked by the try/catch block in gurobi.jl. If the user violates one of the error conditions they cannot see the error msg unless they knows to disable the try/catch block. Is there way to get around that?

lanesmith:
> Hi Lane, I have pushed a new commit that should fix the issues in your comments. One thing I noticed is that the throw error thing is masked by the try/catch block in gurobi.jl. If the user violates one of the error conditions they cannot see the error msg unless they knows to disable the try/catch block. Is there way to get around that?

Hi Dongqi, thanks for taking care of that. I'll be sure to check out the new commit. I believe the problem you're describing is why I used `@error` in conjunction with `throw(ErrorException(""))`. Normally, `throw(ErrorException("some message"))` should be fine, but the error message is masked, as you point out. `@error` seems to work like a fancy print statement, which I believe should still be displayed. That's why, [in this code snippet](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/read.jl#L233), I kept the error message in `@error` and just put `"See above."` as the error message for the `ErrorException`.

danielolsen:
What if we move the `finalize` call to within the `try` block? I.e.:
```julia
try
    global env = Gurobi.Env()
catch e
    println("Error encountered starting Gurobi.")
    throw(e)
end
# Code above here is unchanged
try
    global m = run_scenario(;
        optimizer_factory=env, solver_kwargs=solver_kwargs, kwargs...
    )
    Gurobi.finalize(JuMP.backend(m))
finally
    Gurobi.finalize(env)
    println("Connection to Gurobi closed successfully!")
end
```

I think this handles errors the way we want:
1. If we run into an error setting up the Gurobi environment, a message is printed and the program halts (unchanged).
2. If we run into an error within `run_scenario` (e.g. within `reformat_demand_flexibility_input`, or within the actual solve loop for something like an irreparably infeasible problem), then an error is thrown and the `finally` loop cleans up the Gurobi environment successfully (and we don't need to clean up `m` because it never gets assigned).
3. If everything goes smoothly within `run_scenario`, the JuMP cleanup is done properly within the `try` block and the Gurobi environment cleanup is done properly within the `finally` block.

lanesmith:
> What if we move the `finalize` call to within the `try` block? I.e.:
> 
> ```julia
> try
>     global env = Gurobi.Env()
> catch e
>     println("Error encountered starting Gurobi.")
>     throw(e)
> end
> # Code above here is unchanged
> try
>     global m = run_scenario(;
>         optimizer_factory=env, solver_kwargs=solver_kwargs, kwargs...
>     )
>     Gurobi.finalize(JuMP.backend(m))
> finally
>     Gurobi.finalize(env)
>     println("Connection to Gurobi closed successfully!")
> end
> ```
> 
> I think this handles errors the way we want:
> 
> 1. If we run into an error setting up the Gurobi environment, a message is printed and the program halts (unchanged).
> 2. If we run into an error within `run_scenario` (e.g. within `reformat_demand_flexibility_input`, or within the actual solve loop for something like an irreparably infeasible problem), then an error is thrown and the `finally` loop cleans up the Gurobi environment successfully (and we don't need to clean up `m` because it never gets assigned).
> 3. If everything goes smoothly within `run_scenario`, the JuMP cleanup is done properly within the `try` block and the Gurobi environment cleanup is done properly within the `finally` block.

This sounds good to me. In the event that errors are encountered, this implementation seems like it would have the added benefit of making the resulting stack traces more relevant to the actual problem that is occurred.

dongqi-wu:
> > What if we move the `finalize` call to within the `try` block? I.e.:
> > ```julia
> > try
> >     global env = Gurobi.Env()
> > catch e
> >     println("Error encountered starting Gurobi.")
> >     throw(e)
> > end
> > # Code above here is unchanged
> > try
> >     global m = run_scenario(;
> >         optimizer_factory=env, solver_kwargs=solver_kwargs, kwargs...
> >     )
> >     Gurobi.finalize(JuMP.backend(m))
> > finally
> >     Gurobi.finalize(env)
> >     println("Connection to Gurobi closed successfully!")
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this handles errors the way we want:
> > 
> > 1. If we run into an error setting up the Gurobi environment, a message is printed and the program halts (unchanged).
> > 2. If we run into an error within `run_scenario` (e.g. within `reformat_demand_flexibility_input`, or within the actual solve loop for something like an irreparably infeasible problem), then an error is thrown and the `finally` loop cleans up the Gurobi environment successfully (and we don't need to clean up `m` because it never gets assigned).
> > 3. If everything goes smoothly within `run_scenario`, the JuMP cleanup is done properly within the `try` block and the Gurobi environment cleanup is done properly within the `finally` block.
> 
> This sounds good to me. In the event that errors are encountered, this implementation seems like it would have the added benefit of making the resulting stack traces more relevant to the actual problem that is occurred.

I have incorporated this in the newest commit. Thank you and @danielolsen!

src/read.jl
```diff
@@ -261,3 +255,225 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                converted_zone_flexibility =
+                    Matrix(fh[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_shares[zone_cols_idx, :]
+
+                # add bus-level columns on top of converted bus-level flexibility matrix
+                flex_bus_idx = zeros(Int64, 0)
+                for i in sets.bus_idx
+                    bus = case.busid[i]
+                    # add specified flexibility to the corresponding bus
+                    if bus in flexible_bus_id
+                        converted_zone_flexibility[:, i] += fh[string(bus)]
```
lanesmith, line 437: I receive the following deprecation warning when running this line:

```julia
┌ Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` instead.
│   caller = top-level scope at none:1
└ @ Core none:1
```

It appears to be an issue with how indexing is handled in `fh[string(bus)]`. This might be worth looking into to prevent future package issues. There are a couple other instances below where this type of indexing occurs as well.

src/read.jl
```diff
@@ -261,3 +255,225 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                end
+                eq_bus_df = DataFrames.DataFrame(converted_zone_cost[:, flex_bus_idx])
+            end
+
+            # add header and datetime index column
+            DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flex_bus_idx]))
+            DataFrames.insertcols!(eq_bus_df, 1, :"UTC Time" => fh["UTC Time"])
+
+            # store to new df object
+            demand_flexibility_updated[field] = eq_bus_df
```
lanesmith, line 476: Have you come across a problem with adding DataFrames to the `demand_flexibility_updated` dictionary during your tests? In testing this function, I've run into problems where the initialized dictionary is created to be of type `Dict{String, Union{Nothing, Integer}}` (i.e., a dictionary with keys that are strings and values that are either nothing or integers). Since this is how the dictionary is initialized, I'm not able to add a DataFrame to the dictionary, either by adding a new key or by updating an already established key. I'm not sure if this is due to a difference in Julia versions or package versions, so I want to check. If this is a common problem, it appears that the work around is by specifying that the dictionary can accept keys that are strings and values that can be any datatype through the following:

```julia
demand_flexibility_updated = Dict{String, Any}(
    "duration" => demand_flexibility.duration,
    "enabled" => demand_flexibility.enabled,
    "interval_balance" => demand_flexibility.interval_balance,
    "rolling_balance" => demand_flexibility.rolling_balance,
    "flex_amt_up" => nothing,
    "flex_amt_dn" => nothing,
    "cost_up" => nothing,
    "cost_dn" => nothing,
)
```

src/read.jl
```diff
@@ -261,3 +255,225 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+
+    # check consistency of flexibility input headers
+    if !all(
+        sort(names(demand_flexibility.flex_amt_up)) .==
+        sort(names(demand_flexibility.flex_amt_dn)),
+    )
+        @error("The flexible bus/load zone specified in the up/down input csvs do not 
+               match. Please check the input files to make sure every flexible bus or
+               load zone has corresponding columns in both flexibility csvs.")
+        throw(ErrorException("See above."))
```
lanesmith: With @danielolsen's suggested change to `gurobi.jl`, is this method for stopping the simulation and producing an error message still needed? If it works as expected, only `throw(ExceptionError("The error message."))` should be needed, right?

src/read.jl
dongqi-wu: what is the version of Julia that you are using? I have Julia 1.5.4 (the homepage of REISE says it requires 1.5). I have not seen either problem you mentioned before..

dongqi-wu: image
image
For example I am printing one the DF field after updating and the output looks fine

lanesmith: Okay, that must be the issue on my end. I'm using Julia 1.6 and haven't encountered this issue (or others) before. Thanks for verifying

dongqi-wu: Actually I updated my Julia a couple of weeks ago and then I found that everything fails.. Then I was kinda forced to rollback to 1.5.

danielolsen: @dongqi-wu and @lanesmith thanks for pointing this out, I've started an issue to coordinate information on the 1.5 vs. 1.6 issue and develop a plan of attack: https://github.com/Breakthrough-Energy/REISE.jl/issues/162

src/read.jl
```diff
@@ -261,3 +256,236 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+    # check consistency of flexibility input headers
+    if !all(
+        sort(names(demand_flexibility.flex_amt_up)) .==
+        sort(names(demand_flexibility.flex_amt_dn)),
+    )
+        throw(
+            ErrorException(
+                "The flexible bus/load zone specified in the up/down " *
+                "input csvs do not match. Please check the input files to make sure " *
+                "every flexible bus orload zone has corresponding columns in both " *
```
lanesmith: `orload` -> `or load`

src/read.jl
```diff
@@ -261,3 +256,236 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                    end
+
+                    # check if flexibility in any zone is less than the sum of buses
+                    for i in 1:flexible_zone_num
+                        if any(x -> x < 0, fh[:, zone_columns_idx[flexible_zone_num] + 1])
+                            throw(
+                                ErrorException(
+                                    "Input ERROR: Total zone flexibility less than sum of bus
+                       flexibility for zone " * string(flexible_zone_id[i]),
+                                ),
```
lanesmith, line 421: This passed the linter? The formatting on the string in the `ErrorException` seems a little odd.

lanesmith (approved):
I left two additional comments: one about a typo and the other about string formatting. Once those are addressed, this looks like it's good to go. Thanks!

src/read.jl
dongqi-wu: This is created by lint.. Before I run the formatter it didn't look like this.

src/read.jl
dongqi-wu: fixed, thanks!

src/read.jl
lanesmith: Thanks for clarifying!

PR #163, opened 2021-11-29
feat: enable (penalized) transmission line constraint violations

### LINKED ISSUE #128: Slack Variables for Branch Flow Constraints
https://github.com/Breakthrough-Energy/REISE.jl/blob/d78ee1f18d85480ec52fae8a0e5a6a86902ac960/src/model.jl#L345

It is suggested that we add the slack variables for branch flow limits constrains.  In industry practice, branch flow limits are soft constraints which can be relaxed in day-ahead unit commitment.

danielolsen:

### Purpose
Add a way for users to enable (penalized) transmission line flow violations from the Julia interface, and to analyze them after the result files are extracted. Closes #128.

### What the code is doing
- In **model.jl**: fixing mistakes the prevented transmission violation slack variables from working properly when we pass `trans_viol_enable=true` to `_build_model`, and simplifying the construction of the JuMP expression for branch flow limits when `trans_viol_enable=false` (default).
- In **types.jl**, **query.jl**, and **save.jl**: adding `trans_viol` to the list of variables that will be extracted from the JuMP problem solution and saved to .mat files.
- In **REISE.jl**: adding a new `model_kwargs` input parameter to allow the user to pass model keyword arguments, overriding default settings as applicable.
- In **extract_data.py**: adding `trans_viol` to the list of variables that will be extracted from the collection of .mat files and saved as pickled dataframes, and refactoring the extraction of optimal variables for brevity.

### Testing
Tested manually, by running a scenario with load shedding and transmission violations enabled by default:
```julia
import Gurobi
import REISE
REISE.run_scenario_gurobi(;
    interval=24,
    n_interval=366,
    start_index=1,
    inputfolder=".",
    model_kwargs = Dict(
        "load_shed_enabled" => true,
        "trans_viol_enabled" => true,
        "trans_viol_penalty" => 10000,
    )
)
```

This scenario was known to be infeasible from the start due to minimum generation limit constraints that were incompatible with the local transmission network. Note: this was an ERCOT grid with no DC lines, so this hasn't been tested with mixed AC & DC networks yet.

Then extracting the outputs in the standard way using the `pyreisejl` command-line interface:
```
python path/to/REISE.jl/pyreisejl/utility/extract_data.py -s 2016-01-01 -e 2016-12-31 -x "path/to/outputs" -k
```

As expected, there's a new file **TRANS_VIOL.pkl** created next to **PG.pkl**, **PF.pkl**, etc.

### Time estimate
30 minutes. The largest changes are fairly straightforward refactors, the rest are smaller bug-fixes or tweaked copies of code elsewhere.

### DIFF
--- a/pyreisejl/utility/extract_data.py
+++ b/pyreisejl/utility/extract_data.py
@@ -68,6 +68,15 @@ def extract_data(results):
     sparse_extraction_vars = {"congu", "congl", "load_shed", "trans_viol"}
     temps = {}
     outputs = {}
+    optional_variables = [
+        {"name": "pf_dcline", "key1": "dcline", "key2": "PF_dcline"},
+        {"name": "storage_pg", "key1": "storage", "key2": "PG"},
+        {"name": "storage_e", "key1": "storage", "key2": "Energy"},
+        {"name": "load_shed", "key1": "load_shed", "key2": "load_shed"},
+        {"name": "load_shift_up", "key1": "flexible_demand", "key2": "load_shift_up"},
+        {"name": "load_shift_dn", "key1": "flexible_demand", "key2": "load_shift_dn"},
+        {"name": "trans_viol", "key1": "trans_viol", "key2": "trans_viol"},
+    ]
 
     tic = time.process_time()
     for i, filename in tqdm(enumerate(results)):
@@ -96,34 +105,12 @@ def extract_data(results):
         temps["congl"] = output_mpc["branch"]["MU_ST"].T
 
         # Extract optional variables (not present in all scenarios)
-        try:
-            temps["pf_dcline"] = output_mpc["dcline"]["PF_dcline"].T
-            extraction_vars |= {"pf_dcline"}
-        except KeyError:
-            pass
-
-        try:
-            temps["storage_pg"] = output_mpc["storage"]["PG"].T
-            temps["storage_e"] = output_mpc["storage"]["Energy"].T
-            extraction_vars |= {"storage_pg", "storage_e"}
-        except KeyError:
-            pass
-        try:
-            temps["load_shed"] = output_mpc["load_shed"]["load_shed"].T
-            extraction_vars |= {"load_shed"}
-        except KeyError:
-            pass
-        try:
-            temps["load_shift_up"] = output_mpc["flexible_demand"]["load_shift_up"].T
-            temps["load_shift_dn"] = output_mpc["flexible_demand"]["load_shift_dn"].T
-            extraction_vars |= {"load_shift_up", "load_shift_dn"}
-        except KeyError:
-            pass
-        try:
-            temps["trans_viol"] = output_mpc["trans_viol"]["trans_viol"].T
-            extraction_vars |= {"trans_viol"}
-        except KeyError:
-            pass
+        for var in optional_variables:
+            try:
+                temps[var["name"]] = output_mpc[var["key1"]][var["key2"]].T
+                extraction_vars.add(var["name"])
+            except KeyError:
+                pass
 
         # Extract which number result currently being processed
         i = result_num(filename)
--- a/pyreisejl/utility/extract_data.py
+++ b/pyreisejl/utility/extract_data.py
@@ -65,7 +65,7 @@ def extract_data(results):
     cost = []
 
     extraction_vars = {"pf", "pg", "lmp", "congu", "congl"}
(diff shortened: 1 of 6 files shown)


src/REISE.jl
```diff
@@ -68,12 +69,18 @@ function run_scenario(;
...
+    default_model_kwargs = Dict(
         "case" => case,
         "storage" => storage,
         "demand_flexibility" => demand_flexibility,
         "interval_length" => interval,
     )
+    if isnothing(model_kwargs)
+        model_kwargs = default_model_kwargs
+    else
+        model_kwargs = merge(default_model_kwargs, model_kwargs)
```
BainanXia, line 82: What does `merge` do? I presume it should work similarly as `dict.update` in Python, right?

danielolsen: Exactly, except since Julia is based around multiple dispatch we can call `merge` on either pairs of `AbstractDict` objects (of which `Dict` is a sub-class) or `NamedTuple` objects, or could define another `merge` method that merges other sorts of objects.

https://docs.julialang.org/en/v1/base/collections/#Base.merge

pyreisejl/utility/extract_data.py
```diff
@@ -65,9 +65,18 @@ def extract_data(results):
...
     temps = {}
     outputs = {}
+    optional_variables = [
+        {"name": "pf_dcline", "key1": "dcline", "key2": "PF_dcline"},
+        {"name": "storage_pg", "key1": "storage", "key2": "PG"},
+        {"name": "storage_e", "key1": "storage", "key2": "Energy"},
+        {"name": "load_shed", "key1": "load_shed", "key2": "load_shed"},
+        {"name": "load_shift_up", "key1": "flexible_demand", "key2": "load_shift_up"},
+        {"name": "load_shift_dn", "key1": "flexible_demand", "key2": "load_shift_dn"},
+        {"name": "trans_viol", "key1": "trans_viol", "key2": "trans_viol"},
```
BainanXia, line 78: Is it possible that we replace `key1` and `key2` with more meaningful names? NVM if this is just due to meaningless mat file structure.

pyreisejl/utility/extract_data.py
```diff
@@ -246,10 +238,11 @@ def _get_outputs_from_converted(matfile):
     }
 
     try:
+        # If DC lines are present in the input file, use their indices
         outputs_id["pf_dcline"] = case.mpc.dclineid
+        outputs_id["trans_viol"] = list(case.mpc.branchid) + list(case.mpc.dclineid)
```
BainanXia, line 243: I was thinking whether we should separate ac lines and dc lines in terms of capacity violation?

src/model.jl
```diff
@@ -470,15 +470,11 @@ function _add_constraints_branch_flow_limits!(
...
+        JuMP.@expression(m, branch_limit_pmax, branch_pmax .+ m[:trans_viol])
     else
-        JuMP.@expression(
-            m, branch_limit_pmin[br in sets.branch_idx, h in hour_idx], branch_pmin[br],
-        )
-        JuMP.@expression(
-            m, branch_limit_pmax[br in sets.branch_idx, h in hour_idx], branch_pmax[br],
-        )
+        JuMP.@expression(m, branch_limit_pmin, repeat(branch_pmin, 1, length(hour_idx)))
+        JuMP.@expression(m, branch_limit_pmax, repeat(branch_pmax, 1, length(hour_idx)))
```
BainanXia, line 477: Nice.

pyreisejl/utility/extract_data.py
danielolsen: These are only used within this function to build the `output_mpc` matfile with the structure we want:
```python
for var in optional_variables:
    try:
        temps[var["name"]] = output_mpc[var["key1"]][var["key2"]].T
        extraction_vars.add(var["name"])
    except KeyError:
        pass
```

pyreisejl/utility/extract_data.py
danielolsen: The pre-existing functionality has the `trans_viol` variable defined over all branches, AC & DC. We could limit it to AC branches only if we want, I just added extraction to match what's already here.

BainanXia (approved):
LGTM. We might separate ac lines and dc lines in terms of branch capacity violation in future PRs and we will update user interfaces for this new parameter in PowerSimData.

PR #165, opened 2021-11-30
fix: make trans_viol extraction index a numpy array

danielolsen:

### Purpose
Correct a bug from #163: when DC lines are present, there will be an AttributeError when we try to call `.tolist()` on a list, rather than a numpy array.

### What the code is doing
Instead of converting the two arrays to lists and adding them together, concatenate them into a numpy array.

### Testing
Tested manually on outputs from the Western interconnect: failed at first, now works.

### Time estimate
5 minutes.


BainanXia (approved):
Thanks.

PR #169, opened 2022-03-11
fix: avoid error when demand flexibility isn't enabled

danielolsen:

### Purpose
Fix a bug which causes an exception when demand flexibility isn't enabled. If it's not, then the `demand_flexibility.flex_amt_up` and `demand_flexibility.flex_amt_dn` attributes are `nothing`, and calling `names(nothing)` results in an exception.

### What the code is doing
Before calling `names` on something that may be `nothing`, we check whether demand flexibility is enabled or not. If it's not enabled, then the second condition within the `if` condition will never be evaluated, so the error will be avoided.

### Testing
Tested manually.

### Time estimate
5 minutes.

### DIFF
--- a/src/REISE.jl
+++ b/src/REISE.jl
@@ -70,7 +70,11 @@ function run_scenario(;
     println("All scenario files loaded!")
     case = reise_data_mods(case; num_segments=num_segments)
     sets = _make_sets(case)
-    demand_flexibility = reformat_demand_flexibility_input(case, demand_flexibility, sets)
+    if demand_flexibility.enabled
+        demand_flexibility = reformat_demand_flexibility_input(
+            case, demand_flexibility, sets
+        )
+    end
     save_input_mat(case, storage, inputfolder, outputfolder)
     # Create final model kwargs by merging defaults with user-specified
     default_model_kwargs = Dict(
--- a/src/read.jl
+++ b/src/read.jl
@@ -290,7 +290,8 @@ function reformat_demand_flexibility_input(
 )
 
     # check consistency of flexibility input headers
-    if !all(
+    if demand_flexibility.enabled &&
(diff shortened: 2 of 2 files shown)


lanesmith (approved):
Looks good to me. Thanks! I do wonder why we have it enter `reformate_demand_flexibility_input` at all if demand flexibility is not enabled. I'm not sure if you would want to add it right now (or it could be done later), but we could probably make it so that `reformate_demand_flexibility_input` is only performed if `demand_flexibility.enabled == true` in `REISE.jl`.

BainanXia:
@lanesmith Yeah, I'm wondering about the same thing. I remember I checked logic like this when reviewing the code and somehow figured out such situation could be filtered out by outer logic.

Also, I'm not sure why the Julia tests are failing.

danielolsen:
@lanesmith I think you're right, that would be the more elegant approach:
```julia
demand_flexibility = reformat_demand_flexibility_input(case, demand_flexibility, sets)
```
Could simply become
```julia
if demand_flexibility.enabled
    demand_flexibility = reformat_demand_flexibility_input(case, demand_flexibility, sets)
end
```

I'll test it and update this branch if it works.

danielolsen:
> @lanesmith Yeah, I'm wondering about the same thing. I remember I checked logic like this when reviewing the code and somehow figured out such situation could be filtered out by outer logic.
> 
> Also, I'm not sure why the Julia tests are failing.

We're using dependency versions which were picked based on compatibility with Julia 1.5.4 at the time, but aren't compatible with Julia 1.7.2. We've known for a little while that there were compatibility issues with newer versions of Julia, I guess this is the first time that something has broken in trying to set up the environment (since we don't currently test much else).

PR #175, opened 2022-03-23
chore: update flask

jenhagg:
### Purpose
Address startup failure in docker container due to outdated dependencies. 
```
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/usr/local/lib/python3.8/site-packages/markupsafe/__init__.py)
```

### What the code is doing
Use newer version of flask which fixes the transitive dependency.

### Testing
Built the docker image and started the container using plug. Checked that web requests are handled successfully.

### Time estimate
1 min

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ pandas~=1.2
 scipy~=1.6
 tqdm~=4.29.1
 pytest~=6.1.1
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
Thanks for the quick fix!

PR #178, opened 2022-04-23
feat: integrate DOE flexibility data and percentage-based input

dongqi-wu:

### Purpose

Integrate pre-processed EIA-ID based flexibility data from DOE flexibility data into REISE modelling.
The DOE data will be downloaded from blob storage first if not found in the case folder.
The EIA ID of buses is stored in column 18 of the bus field in case.mat. This file is manually modified for now, but in the future this info will be integrated into powersimdata.

### What the code is doing
Read the percentage based LSE flexibility, match with load buses and convert to MW amount

### Testing
run_scenario_gurobi() using TX sub-system data

### Where to look
read.jl: additional logic for reading DOE data
model.jl: convert percentage to MW 

### Time estimate
15 min

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -58,18 +58,54 @@ Given a Case object and a DemandFlexibility object, build a matrix of demand fle
     by (bus, hour) for this interval.
 """
 function _make_bus_demand_flexibility_amount(
-    case::Case, demand_flexibility::DemandFlexibility, start_index::Int, end_index::Int
+    case::Case,
+    demand_flexibility::DemandFlexibility,
+    start_index::Int,
+    end_index::Int,
+    bus_demand::Matrix,
+    sets::Sets,
 )::Tuple{Matrix,Matrix}
-    # Bus weighting
-    zone_to_bus_shares = _make_bus_demand_weighting(case)
-    # Demand flexibility up profile
-    simulation_demand_flex_amt_up = Matrix(
-        demand_flexibility.flex_amt_up[start_index:end_index, 2:end]
-    )
-    # Demand flexibility down profile
-    simulation_demand_flex_amt_dn = Matrix(
-        demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
-    )
+
+    # combine two profiles with user input overriding DOE flexibility
+    simulation_demand_flex_amt_up = zeros(end_index - start_index + 1, sets.num_bus)
+    simulation_demand_flex_amt_dn = zeros(end_index - start_index + 1, sets.num_bus)
+
+    # DOE flexibility, convert to MW, DOE timestamps already aligned
+    if demand_flexibility.enable_doe_flexibility
+        doe_flex_pct = Matrix(demand_flexibility.doe_flex_amt[start_index:end_index, 2:end])
+        doe_flex_mw = doe_flex_pct .* permutedims(bus_demand[sets.doe_flexible_bus_idx, :])
+        for i in 1:length(sets.doe_flexible_bus_idx)
+            simulation_demand_flex_amt_up[:, sets.doe_flexible_bus_idx[i]] = doe_flex_mw[
+                :, i
+            ]
+            simulation_demand_flex_amt_dn[:, sets.doe_flexible_bus_idx[i]] = doe_flex_mw[
+                :, i
+            ]
+        end
+    end
+
+    # replace columns with user input numbers if applicable
+    if !isnothing(demand_flexibility.flex_amt_up)
+        csv_flex_amt_up = Matrix(
+            demand_flexibility.flex_amt_up[start_index:end_index, 2:end]
+        )
+        csv_flex_amt_dn = Matrix(
+            demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
+        )
+
+        for i in 1:length(sets.csv_flexible_bus_idx)
+            simulation_demand_flex_amt_up[:, sets.csv_flexible_bus_idx[i]] = csv_flex_amt_up[
+                :, i
+            ]
+            simulation_demand_flex_amt_dn[:, sets.csv_flexible_bus_idx[i]] = csv_flex_amt_dn[
+                :, i
+            ]
+        end
+    end
+
+    # remove non-flexible columns
+    simulation_demand_flex_amt_up = simulation_demand_flex_amt_up[:, sets.flexible_bus_idx]
+    simulation_demand_flex_amt_dn = simulation_demand_flex_amt_dn[:, sets.flexible_bus_idx]
 
     return (
         permutedims(simulation_demand_flex_amt_up),
@@ -146,15 +182,32 @@ function _make_sets(
     demand_flexibility_enabled =
         isa(demand_flexibility, DemandFlexibility) && demand_flexibility.enabled
     if demand_flexibility_enabled
+        # if DOE profile is used, all buses with valid EIA ID are flexible
+        if demand_flexibility.enable_doe_flexibility
+            doe_flexible_bus_idx = sort(
+                intersect(load_bus_idx, findall(case.bus_eiaid .> 0))
+            )
+        else
+            doe_flexible_bus_idx = nothing
+        end
+
+        # flexible buses from input files    
         # assume each flexible bus can go up/dn, so only use the Dataframe for up
-        flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
-        flexible_bus_id = [parse(Int64, bus) for bus in flexible_bus_str]
-        flexible_bus_idx = [bus_id2idx[bus] for bus in flexible_bus_id]
+        csv_flexible_bus_str = names(demand_flexibility.flex_amt_up)[2:end]
+        csv_flexible_bus_id = [parse(Int64, bus) for bus in csv_flexible_bus_str]
+        csv_flexible_bus_idx = [bus_id2idx[bus] for bus in csv_flexible_bus_id]
+
+        # all flexible buses
+        flexible_bus_idx = sort([
+            i for i in union(doe_flexible_bus_idx, csv_flexible_bus_idx)
+        ])
         num_flexible_bus = length(flexible_bus_idx)
         flexible_load_bus_map = sparse(
             flexible_bus_idx, 1:num_flexible_bus, 1, num_bus, num_flexible_bus
         )::SparseMatrixCSC
     else
+        csv_flexible_bus_idx = nothing
+        doe_flexible_bus_idx = nothing
         flexible_bus_idx = nothing
         num_flexible_bus = 0
         flexible_load_bus_map = nothing
@@ -190,6 +243,8 @@ function _make_sets(
         segment_idx=segment_idx,
         num_storage=num_storage,
         storage_idx=storage_idx,
+        csv_flexible_bus_idx=csv_flexible_bus_idx,
+        doe_flexible_bus_idx=doe_flexible_bus_idx,
         flexible_bus_idx=flexible_bus_idx,
         num_flexible_bus=num_flexible_bus,
         flexible_load_bus_map=flexible_load_bus_map,
@@ -604,7 +659,7 @@ function _build_model(
     # Demand flexibility parameters (if present)
     if demand_flexibility.enabled
         (bus_demand_flex_amt_up, bus_demand_flex_amt_dn) = _make_bus_demand_flexibility_amount(
-            case, demand_flexibility, start_index, end_index
+            case, demand_flexibility, start_index, end_index, bus_demand, sets
         )
     end
 
--- a/src/model.jl
+++ b/src/model.jl
@@ -198,9 +198,13 @@ function _make_sets(
         csv_flexible_bus_idx = [bus_id2idx[bus] for bus in csv_flexible_bus_id]
 
         # all flexible buses
-        flexible_bus_idx = sort([
-            i for i in union(doe_flexible_bus_idx, csv_flexible_bus_idx)
-        ])
+        if !isnothing(doe_flexible_bus_idx)
+            flexible_bus_idx = sort([
+                i for i in union(doe_flexible_bus_idx, csv_flexible_bus_idx)
+            ])
+        else
+            flexible_bus_idx = sort(csv_flexible_bus_idx)
+        end
         num_flexible_bus = length(flexible_bus_idx)
         flexible_load_bus_map = sparse(
             flexible_bus_idx, 1:num_flexible_bus, 1, num_bus, num_flexible_bus
--- a/src/read.jl
+++ b/src/read.jl
@@ -26,6 +26,11 @@ function read_case(filepath)
     case["busid"] = convert(Array{Int,1}, bus.bus_id)
     case["bus_demand"] = convert(Array{Float64,1}, bus.Pd)
     case["bus_zone"] = convert(Array{Int,1}, bus.zone_id)
+    try
+        case["bus_eiaid"] = convert(Array{Int,1}, bus.eia_id)
+    catch e
+        case["bus_eiaid"] = zeros(size(case["busid"], 1), 1)
+    end
 
     # Generators
     plant = CSV.File(joinpath(filepath, "plant.csv"))
@@ -105,6 +110,7 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         "enabled" => "not_specified",
         "interval_balance" => true,
         "rolling_balance" => true,
+        "enable_doe_flexibility" => false,
     )
 
     # Try loading the demand flexibility parameters
@@ -133,6 +139,10 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
             "rolling_balance" => (
                 "The parameter that indicates if the rolling load balance constraint " *
                 "is enabled is not defined. Will default to being enabled."
+            ),
+            "enable_doe_flexibility" => (
+                "The parameter that indicates if the DOE flexibility data will be used" *
+                " to parameterize demand flexibility profiles. Will default to disabled."
             ),
         )
 
@@ -165,6 +175,33 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     # Prevent the rolling_balance constraint according to the duration parameter
     demand_flexibility["rolling_balance"] &= !(demand_flexibility["duration"] == interval)
 
+    # Try loading DOE flexibility profile if enabled
+    demand_flexibility["doe_flex_amt"] = nothing
+    if demand_flexibility["enable_doe_flexibility"] == true && demand_flexibility["enabled"]
+        try
+            # check if DOE data is present, if not, download from BLOB server
+            if !isfile(joinpath(filepath, "doe_flexibility_2016.csv"))
+                println(
+                    "DOE flexibility data is enabled, but a local copy " *
+                    "is not present. Downloading data from BLOB storage..",
+                )
+                @sync download(
+                    "https://besciences.blob.core.windows.net/datasets/" *
+                    "demand_flexibility_doe/doe_flexibility_2016.csv",
+                    joinpath(filepath, "doe_flexibility_2016.csv"),
+                )
+                println("Successfully downloaded DOE flexibility file.")
+            end
+            # read local file
+            demand_flexibility["doe_flex_amt"] = DataFrames.DataFrame(
+                CSV.File(joinpath(filepath, "doe_flexibility_2016.csv"))
+            )
+            println("...loading DOE demand flexibility profiles")
+        catch e
+            println("DOE demand flexibility profile not found on BLOB storage")
+        end
+    end
+
     # Try loading the demand flexibility and demand flexibility cost profiles
     for s in ["up", "dn"]
         # Pre-specify the demand flexibility and demand flexibility cost profiles
@@ -174,13 +211,12 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         # Only try loading the profiles if demand flexibility is enabled
         if demand_flexibility["enabled"] == "not_specified" ||
             (demand_flexibility["enabled"])
-            # Try loading the demand flexibility profiles
             try
                 demand_flexibility["flex_amt_" * s] = DataFrames.DataFrame(
                     CSV.File(joinpath(filepath, "demand_flexibility_" * s * ".csv"))
                 )
                 println("...loading demand flexibility " * s * " profiles")
-            catch
+            catch e
                 println("Demand flexibility " * s * " profile not found in " * filepath)
             end
 
@@ -190,7 +226,7 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
                     CSV.File(joinpath(filepath, "demand_flexibility_cost_" * s * ".csv"))
                 )
                 println("...loading demand flexibility " * s * "-shift cost profiles")
-            catch
+            catch e
                 println(
                     "Demand flexibility " *
                     s *
@@ -205,10 +241,12 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     end
 
     # If demand flexibility is enabled but at least one demand flexibility profile is nothing
-    if demand_flexibility["enabled"] == true && (
-        isnothing(demand_flexibility["flex_amt_up"]) ||
-        (isnothing(demand_flexibility["flex_amt_dn"]))
-    )
+    if demand_flexibility["enabled"] == true &&
+        (
+            isnothing(demand_flexibility["flex_amt_up"]) ||
+            (isnothing(demand_flexibility["flex_amt_dn"]))
+        ) &&
+        isnothing(demand_flexibility["doe_flex_amt"])
         throw(
             ErrorException(
                 "Demand flexibility was specified to be enabled, however at " *
@@ -219,7 +257,8 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         )
     elseif demand_flexibility["enabled"] == "not_specified"
         if !isnothing(demand_flexibility["flex_amt_up"]) &&
-            (!isnothing(demand_flexibility["flex_amt_dn"]))
+           (!isnothing(demand_flexibility["flex_amt_dn"])) ||
+            !isnothing(demand_flexibility["doe_flex_amt"])
             demand_flexibility["enabled"] = true
         else
             if !isnothing(demand_flexibility["flex_amt_up"]) ||
@@ -319,20 +358,22 @@ function reformat_demand_flexibility_input(
         "flex_amt_dn" => nothing,
         "cost_up" => nothing,
         "cost_dn" => nothing,
+        "doe_flex_amt" => nothing,
+        "enable_doe_flexibility" => demand_flexibility.enable_doe_flexibility,
     )
 
     # iterate through fields
     for field in ["flex_amt_up", "flex_amt_dn", "cost_up", "cost_dn"]
         # dataframe in the corresponding field
-        fh = getfield(demand_flexibility, Symbol(field))
+        demand_flex_field = getfield(demand_flexibility, Symbol(field))
 
         # skip if no cost
-        if isnothing(fh)
+        if isnothing(demand_flex_field)
             continue
         end
 
         # headers
-        flexible_str = names(fh)[2:end]
+        flexible_str = names(demand_flex_field)[2:end]
 
         # index of columns specifing the flexibility of a zone
         zone_columns_idx = findall(x -> occursin("zone.", x), flexible_str)
@@ -396,7 +437,7 @@ function reformat_demand_flexibility_input(
                     for i in 1:flexible_bus_num
                         # substract from total if the zone of this bus is specified in a zone column
                         if !isnothing(bus_cols_zone_idx[i])
-                            fh[:, zone_columns_idx[bus_cols_zone_idx[i]] + 1] -= fh[
+                            demand_flex_field[:, zone_columns_idx[bus_cols_zone_idx[i]] + 1] -= demand_flex_field[
                                 :, bus_columns_idx[i] + 1
                             ]
                         end
@@ -404,7 +445,10 @@ function reformat_demand_flexibility_input(
 
                     # check if flexibility in any zone is less than the sum of buses
                     for i in 1:flexible_zone_num
-                        if any(x -> x < 0, fh[:, zone_columns_idx[flexible_zone_num] + 1])
+                        if any(
+                            x -> x < 0,
+                            demand_flex_field[:, zone_columns_idx[flexible_zone_num] + 1],
+                        )
                             throw(
                                 ErrorException(
                                     "Input ERROR: Total zone flexibility less than sum of bus
@@ -417,7 +461,7 @@ function reformat_demand_flexibility_input(
 
                 # convert zone-level to bus-level
                 converted_zone_flexibility =
-                    Matrix(fh[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_shares[zone_cols_idx, :]
+                    Matrix(demand_flex_field[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_shares[zone_cols_idx, :]
 
                 # add bus-level columns on top of converted bus-level flexibility matrix
                 flex_bus_idx = zeros(Int64, 0)
@@ -425,7 +469,9 @@ function reformat_demand_flexibility_input(
                     bus = case.busid[i]
                     # add specified flexibility to the corresponding bus
                     if bus in flexible_bus_id
-                        converted_zone_flexibility[:, i] += fh[string(bus)]
+                        converted_zone_flexibility[:, i] += demand_flex_field[
+                            !, string(bus)
+                        ]
                     end
 
                     # identify flexible bus by their total flexibility and append to list
@@ -434,13 +480,13 @@ function reformat_demand_flexibility_input(
                     end
                 end
                 eq_bus_df = DataFrames.DataFrame(
-                    converted_zone_flexibility[:, flex_bus_idx]
+                    converted_zone_flexibility[:, flex_bus_idx], :auto
                 )
                 # for cost, the zone numbers apply to all buses except those with dedicated columns
             else
                 # convert zone-level cost using incidence matrix
                 converted_zone_cost =
-                    Matrix(fh[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_incidence[zone_cols_idx, :]
+                    Matrix(demand_flex_field[:, [i + 1 for i in zone_columns_idx]]) * zone_to_bus_incidence[zone_cols_idx, :]
 
                 # replace bus-level cost for bus columns in converted bus-level cost matrix
                 flex_bus_idx = zeros(Int64, 0)
@@ -448,7 +494,7 @@ function reformat_demand_flexibility_input(
                     bus = case.busid[i]
                     # add specified cost to the corresponding bus
                     if bus in flexible_bus_id
-                        converted_zone_cost[:, i] = fh[string(bus)]
+                        converted_zone_cost[:, i] = demand_flex_field[!, string(bus)]
                     end
 
                     # identify flexible bus by their total cost and append to list
@@ -456,20 +502,79 @@ function reformat_demand_flexibility_input(
                         append!(flex_bus_idx, i)
                     end
                 end
-                eq_bus_df = DataFrames.DataFrame(converted_zone_cost[:, flex_bus_idx])
+
+                # if DOE flexibility is used, add columns for flexible buses there
+
+                # find all flexible buses
+                if demand_flexibility.enable_doe_flexibility
+                    doe_flexible_bus_idx = sort(
+                        intersect(sets.load_bus_idx, findall(case.bus_eiaid .> 0))
+                    )
+                else
+                    doe_flexible_bus_idx = nothing
+                end
+
+                # assume each flexible bus can go up/dn, so only use the Dataframe for up
+                csv_flexible_bus_str = names(demand_flexibility_updated["flex_amt_up"])[2:end]
+                csv_flexible_bus_id = [parse(Int64, bus) for bus in csv_flexible_bus_str]
+                csv_flexible_bus_idx = [sets.bus_id2idx[bus] for bus in csv_flexible_bus_id]
+
+                # all flexible buses
+                flex_bus_idx = sort([
+                    i for i in union(doe_flexible_bus_idx, csv_flexible_bus_idx)
+                ])
+
+                new_demand_flex_cost = zeros(size(converted_zone_cost, 1), sets.num_bus)
+                for i in 1:length(flex_bus_idx)
+                    new_demand_flex_cost[:, flex_bus_idx[i]] = converted_zone_cost[:, i]
+                end
+
+                eq_bus_df = DataFrames.DataFrame(
+                    new_demand_flex_cost[:, flex_bus_idx], :auto
+                )
             end
 
             # add header and datetime index column
             DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flex_bus_idx]))
-            DataFrames.insertcols!(eq_bus_df, 1, :"UTC Time" => fh["UTC Time"])
+            DataFrames.insertcols!(
+                eq_bus_df, 1, :"UTC Time" => demand_flex_field[!, "UTC Time"]
+            )
 
             # store to new df object
             demand_flexibility_updated[field] = eq_bus_df
 
             # if all columns are buses, use the original dataframe
         else
-            demand_flexibility_updated[field] = fh
+            demand_flexibility_updated[field] = demand_flex_field
         end
+    end
+
+    # re-format DOE flexibility using bus EIA ID
+    if demand_flexibility.enable_doe_flexibility == true
+        demand_flex_field = getfield(demand_flexibility, Symbol("doe_flex_amt"))
+
+        # all load buses with valid EIA ID 
+        flexible_bus_idx = intersect(sets.load_bus_idx, findall(case.bus_eiaid .> 0))
+        flexible_bus_num = length(flexible_bus_idx)
+
+        # bus flexibility Percentage matrix
+        doe_bus_flexibility = zeros(size(case.demand, 1), flexible_bus_num)
+        doe_eiaid = [parse(Int, x) for x in names(demand_flex_field)[2:end]]
+        for i in 1:flexible_bus_num
+            doe_bus_flexibility[:, i] = demand_flex_field[
+                !, Symbol(case.bus_eiaid[flexible_bus_idx[i]])
+            ]
+        end
+
+        eq_bus_df = DataFrames.DataFrame(doe_bus_flexibility, :auto)
+        # add header and datetime index column
+        DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flexible_bus_idx]))
+        DataFrames.insertcols!(
+            eq_bus_df, 1, :"UTC Time" => demand_flex_field[!, "UTC Time"]
+        )
+
+        # store to new df object
+        demand_flexibility_updated["doe_flex_amt"] = eq_bus_df
     end
 
     # Convert Dict to type DemandFlexibility
--- a/src/read.jl
+++ b/src/read.jl
@@ -175,33 +175,6 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     # Prevent the rolling_balance constraint according to the duration parameter
     demand_flexibility["rolling_balance"] &= !(demand_flexibility["duration"] == interval)
 
-    # Try loading DOE flexibility profile if enabled
-    demand_flexibility["doe_flex_amt"] = nothing
-    if demand_flexibility["enable_doe_flexibility"] == true && demand_flexibility["enabled"]
-        try
-            # check if DOE data is present, if not, download from BLOB server
-            if !isfile(joinpath(filepath, "doe_flexibility_2016.csv"))
-                println(
-                    "DOE flexibility data is enabled, but a local copy " *
-                    "is not present. Downloading data from BLOB storage..",
-                )
-                @sync download(
-                    "https://besciences.blob.core.windows.net/datasets/" *
-                    "demand_flexibility_doe/doe_flexibility_2016.csv",
-                    joinpath(filepath, "doe_flexibility_2016.csv"),
-                )
-                println("Successfully downloaded DOE flexibility file.")
-            end
-            # read local file
-            demand_flexibility["doe_flex_amt"] = DataFrames.DataFrame(
-                CSV.File(joinpath(filepath, "doe_flexibility_2016.csv"))
-            )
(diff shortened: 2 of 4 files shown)


src/read.jl
```diff
@@ -41,6 +41,7 @@ function read_case(filepath)
     case["busid"] = convert(Array{Int,1}, mpc["bus"][:, 1])
     case["bus_demand"] = mpc["bus"][:, 3]
     case["bus_zone"] = convert(Array{Int,1}, mpc["bus"][:, 7])
+    case["bus_eiaid"] = convert(Array{Int,1}, mpc["bus"][:, 18])
```
danielolsen: The idea is that this new column will be added to the base grid and/or add Scenario creation time, yeah?

dongqi-wu: > The idea is that this new column will be added to the base grid and/or add Scenario creation time, yeah?

right

src/read.jl
```diff
@@ -114,6 +115,7 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         "enabled" => "not_specified",
         "interval_balance" => true,
         "rolling_balance" => true,
+        "source" => "doe",
```
danielolsen: It seems like this `source` parameter controls a couple of things if it's set to `"doe"`:
- the filepath that flexibility information is loaded from
- that bus flexibility needs to come from mapping each bus's EIA ID to the corresponding column in the loaded CSV
- that flexibility is interpreted as a percentage of demand, not absolute MW
- others?

Are there situations where we may want to be doing some of these things, but not others? If so, we might want to make this control a bit more granular.

src/read.jl
```diff
@@ -475,6 +500,31 @@ function reformat_demand_flexibility_input(
...
+            flexible_bus_idx = intersect(sets.load_bus_idx, findall(case.bus_eiaid .> 0))
+            flexible_bus_num = length(flexible_bus_idx)
+            
+            # bus flexibility Percentage matrix
+            doe_bus_flexibility = zeros(size(case.demand, 1), flexible_bus_num)
+            doe_eiaid = [parse(Int, x) for x in names(fh)[2:end]]
+            for i in 1:flexible_bus_num
+                if case.bus_eiaid[flexible_bus_idx[i]] in doe_eiaid
+                    doe_bus_flexibility[:, i] = fh[Symbol(case.bus_eiaid[flexible_bus_idx[i]])]
+                end
```
danielolsen: Do we want to warn the user somehow if a bus has an EIA ID that doesn't have a corresponding entry in the flexibility data?

src/read.jl
dongqi-wu: 1. The filepath is a temporary thing for now because ideally this file will be on the blob server.
2. yes, it is assumed the case file have an EIA for each bus, which will be done outside REISE
3. right 
now it is either from user specified input csvs (like before) or from DOE data so it only does one of the 2 things. If it is to be generalized there would need a few more parameters.

src/read.jl
```diff
@@ -475,6 +500,31 @@ function reformat_demand_flexibility_input(
...
+        elseif demand_flexibility.source == "doe" &&
+                field in ["flex_amt_up", "flex_amt_dn"]
+            
+            # all load buses with valid EIA ID 
+            flexible_bus_idx = intersect(sets.load_bus_idx, findall(case.bus_eiaid .> 0))
+            flexible_bus_num = length(flexible_bus_idx)
+            
+            # bus flexibility Percentage matrix
+            doe_bus_flexibility = zeros(size(case.demand, 1), flexible_bus_num)
+            doe_eiaid = [parse(Int, x) for x in names(fh)[2:end]]
```
danielolsen: I know this was introduced before this PR, but can we change the name of the `fh` variable to something a little more meaningful?

dongqi-wu: sure

src/read.jl
dongqi-wu: Yes, a check will be added later. In the Texas subsystem I am using now there are no such buses so it does not prompt an error, but this will be a problem for the USA network. There are like 11 load buses in the USA system that are located in Canada and Mexico that do not have an EIAID associated with them. They are marked as 0 for now and in the future we might want to assign them to their nearest neighbor.

src/read.jl
danielolsen: We'll probably want to have some `try`/`catch` logic around this line so that if there's no 18th column in the bus table, `case["bus_eiaid"]` gets filled with zeros rather than throwing an error.

src/read.jl
```diff
@@ -183,16 +191,32 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                    )
+                    println("...loading DOE demand flexibility " * s * " profiles")
+                catch e
+                    println("Demand flexibility " * s * " profile not found in " * filepath)
+                end               
+            # throw an error otherwise
+            else
+                demand_flexibility["flex_amt_" * s] = nothing
+                println("Demand flexibility parameter" * demand_flexibility["source"] *
+                        "is invalid!")
```
lanesmith: In this case, might it be best to throw an error that stops the simulation? Perhaps by doing something like `throw(ErrorException("The provided demand flexibility source is invalid"))`? Ideally, `PowerSimData` will (eventually) catch something like this, but it's always good for the standalone version of `REISE.jl` to catch this as well.

src/read.jl
```diff
@@ -143,6 +145,10 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
                 "The parameter that indicates if the rolling load balance constraint " *
                 "is enabled is not defined. Will default to being enabled."
             ),
+            "source" => (
+                "The parameter that indicates the source of demand flexibility amount " *
+                "is not defined. Will default to DOE data."
+            )
```
lanesmith: I know the current implementation involves passing the DOE data into `REISE.jl` via profiles, but once we go to the planned method where the DOE data is either in blob storage or hosted locally, we might want to rethink this default. For instance, if the source parameter is not provided but demand flexibility profiles are provided, then the source should probably be set to `"user"`. Alternatively, if the source parameter is not provided, demand flexibility profiles are not provided, and demand flexibility is enabled, then the source should probably be set to `"doe"`.

EDIT: Something like this shouldn't be a problem once we get the `PowerSimData` interface updated to handle these new demand flexibility inputs, but it will be good to have this for the standalone version of `REISE.jl`.

dongqi-wu:
Most of the comments have been addressed in the recent commit. This version lets the user to choose flexibility input between using DOE data (downloaded from the BLOB storage) or local csv files. The code assumes the EIA ID of bus ([feat: EIA ID of buses in the synthetic grid #623](https://github.com/Breakthrough-Energy/PowerSimData/pull/623)) is present in the .mat case file.

src/read.jl
```diff
@@ -183,16 +193,51 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
-            try
-                demand_flexibility["flex_amt_" * s] = DataFrames.DataFrame(
-                    CSV.File(joinpath(filepath, "demand_flexibility_" * s * ".csv"))
+            # Try loading the DOE data from BLOB storgae
+            if demand_flexibility["enable_doe_flexibility"]
+                try
+                    # check if DOE data is present, if not, download from BLOB server
+                    if !isfile(joinpath(filepath, "doe_flexibility_2016.csv"))
+                        println(
+                            "DOE flexibility data is enabled, but a local copy" *
```
lanesmith: Add a space between `copy` and `"`

src/read.jl
```diff
@@ -183,16 +193,51 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+                    end
+                    # read local file
+                    demand_flexibility["flex_amt_" * s] = DataFrames.DataFrame(
+                        CSV.File(joinpath(filepath, "doe_flexibility_2016.csv"))
+                    )
+                    println("...loading DOE demand flexibility profiles")
+                catch e
+                    println("DOE demand flexibility profile not found on BLOB storage")
+                end
+                # Load from user defined demand flexibility data if DOE disabled
```
lanesmith: Unindent this comment one level (i.e., it should be in line with the following `elseif` statement)

src/read.jl
```diff
@@ -183,16 +193,51 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
...
+            elseif demand_flexibility["enable_doe_flexibility"] == false
+                try
+                    demand_flexibility["flex_amt_" * s] = DataFrames.DataFrame(
+                        CSV.File(joinpath(filepath, "demand_flexibility_" * s * ".csv"))
+                    )
+                    println("...loading demand flexibility " * s * " profiles")
+                catch e
+                    println("Demand flexibility " * s * " profile not found in " * filepath)
+                end
+                # throw an error otherwise
```
lanesmith: Unindent this comment one level (i.e., it should be in line with the following `else` statement)

src/read.jl
```diff
@@ -470,14 +519,45 @@ function reformat_demand_flexibility_input(
...
+            # add header and datetime index column
+            DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flexible_bus_idx]))
+            DataFrames.insertcols!(
+                eq_bus_df, 1, :"UTC Time" => demand_flex_field["UTC Time"]
+            )
 
             # store to new df object
             demand_flexibility_updated[field] = eq_bus_df
 
             # if all columns are buses, use the original dataframe
```
lanesmith, line 554: Unindent this comment one level (i.e., it should be in line with the following `else` statement)

src/read.jl
```diff
@@ -470,14 +519,45 @@ function reformat_demand_flexibility_input(
...
+            # bus flexibility Percentage matrix
+            doe_bus_flexibility = zeros(size(case.demand, 1), flexible_bus_num)
+            doe_eiaid = [parse(Int, x) for x in names(demand_flex_field)[2:end]]
+            for i in 1:flexible_bus_num
+                if case.bus_eiaid[flexible_bus_idx[i]] in doe_eiaid
+                    doe_bus_flexibility[:, i] = demand_flex_field[Symbol(
+                        case.bus_eiaid[flexible_bus_idx[i]]
+                    )]
+                end
+            end
```
lanesmith: Has this been tested for the larger grids? I'm just curious if the step where demand flexibility gets read and reformatted takes a while because iterating through all demand flexibility buses seems like it could be slow.

src/model.jl
```diff
@@ -71,6 +75,12 @@ function _make_bus_demand_flexibility_amount(
         demand_flexibility.flex_amt_dn[start_index:end_index, 2:end]
     )
 
+    # percentage flexibility from input
+    if demand_flexibility.enable_doe_flexibility
+        simulation_demand_flex_amt_up .*= permutedims(bus_demand[sets.flexible_bus_idx, :])
+        simulation_demand_flex_amt_dn .*= permutedims(bus_demand[sets.flexible_bus_idx, :])
+    end
```
lanesmith: This is fine for this PR, but we might need to rethink this for the forthcoming PR we've previously discussed that will allow for the mixed input of profiles (with MW values) and EIA data (with percentage values). In such a setup, we'll probably need separate variables for holding profile-specific data and EIA-specific data, whereas right now we're able to use `demand_flexibility.flex_amt_up` and `demand_flexibility.flex_amt_dn` for both situations.

lanesmith (commented):
It looks pretty good. See my comments below.

src/model.jl
dongqi-wu: correct, I am storing the DOE flexibility data under new fields in the next version. Currently the user has to choose between DOE or input csvs so when using the DOE flexibility it also uses the old container fields.

src/read.jl
dongqi-wu: I guess the if check might be slow when it checks if the eiaid of each bus is valid. Maybe this can be removed as the check is already done when assigning the eiaid to buses in prereise.

src/read.jl
dongqi-wu: It seems JuliaFormatter prefers this way. I don't remember writing like this and it was done by the formatter. I tried to do as you suggested but the last commit was rejected by Lint, and I ran the formatter again it changed the comments back to this version.

lanesmith: Thanks for letting me know!

src/model.jl
lanesmith: Okay, sounds good!

src/read.jl
lanesmith: Since we are providing that data and the check is already done in `PreREISE`, I would agree that having the check in `REISE.jl` is probably not necessary. @BainanXia, what are your thoughts? Might a check like this still be necessary if users are adding extra buses? Or would this check make more sense in `PowerSimData` (once we add this to `PowerSimData`) for new buses?

src/read.jl
```diff
@@ -143,6 +149,10 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
                 "The parameter that indicates if the rolling load balance constraint " *
                 "is enabled is not defined. Will default to being enabled."
             ),
+            "enable_doe_flexibility" => (
+                "The parameter that the DOE flexibility data will be used as baseline" *
+                "in the simulation."
```
lanesmith: Maybe change this to `"The parameter that indicates if the DOE flexibility data will be used to parameterize the demand flexibility profiles. Will default to not being enabled."`? If so, be sure to split the string to meet the desired line length.

src/read.jl
```diff
@@ -183,16 +193,51 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
         # Only try loading the profiles if demand flexibility is enabled
         if demand_flexibility["enabled"] == "not_specified" ||
             (demand_flexibility["enabled"])
-            # Try loading the demand flexibility profiles
-            try
-                demand_flexibility["flex_amt_" * s] = DataFrames.DataFrame(
-                    CSV.File(joinpath(filepath, "demand_flexibility_" * s * ".csv"))
+            # Try loading the DOE data from BLOB storgae
+            if demand_flexibility["enable_doe_flexibility"]
```
lanesmith: Since this conditional block is in part checking if the `enable_doe_flexibility` parameter is a boolean, I believe this first conditional check will need to be `if demand_flexibility["enable_doe_flexibility"] == true`. If the `enable_doe_flexibility` parameter is not a boolean, then there would be an error because a non-boolean would be trying to be used in a boolean context.

src/read.jl
```diff
@@ -470,14 +519,45 @@ function reformat_demand_flexibility_input(
...
             DataFrames.rename!(eq_bus_df, Symbol.(case.busid[flex_bus_idx]))
-            DataFrames.insertcols!(eq_bus_df, 1, :"UTC Time" => fh["UTC Time"])
+            DataFrames.insertcols!(
+                eq_bus_df, 1, :"UTC Time" => demand_flex_field["UTC Time"]
+            )
+
+            # store to new df object
+            demand_flexibility_updated[field] = eq_bus_df
+
+            # if flex amount specified in DOE data, re-format using bus EIA ID
```
lanesmith: Unindent one level to make this comment inline with the `elseif` conditional

lanesmith (commented):
I think this mostly looks good to me. Please see my comments below.

Additionally, do we have a plan for mixing user-provided profiles (i.e., the original method) and the profiles derived from EIA data? I remember we talked about it before, but it's been a while.

src/read.jl
lanesmith: I just wanted to follow-up on my previous comment here. Thoughts?

dongqi-wu:
> I think this mostly looks good to me. Please see my comments below.
> 
> Additionally, do we have a plan for mixing user-provided profiles (i.e., the original method) and the profiles derived from EIA data? I remember we talked about it before, but it's been a while.

The plan was to make the model take both, with the csv files overriding the EIA ones if needed. This will be the next version.

src/read.jl
dongqi-wu: Actually this seems to be the way the formatter loves

dongqi-wu:
I just pushed another commit to enable using both DOE and csv flexibility, while csv columns overriding the defaults ones from DOE data. This requires modifying the  `demand_flexibility.cost_up/dn` after reading `flex_cost_up/dn.csv` into the `demand_flexibility` struct, as the total number of flexible bus will be different than the number of columns in the csvs when DOE data is used, and the dimension of the DF need to be modified accordingly. There might be cleaner ways but for this version I just want the logic to be reviewed.

Also the test shows 
> ERROR: Could not find either `git-tree-sha1` or `path` for package DelimitedFiles

idk if its my problem or something with the pip equivalent for Julia, please let me know if you have seen this before.

lanesmith:
> I just pushed another commit to enable using both DOE and csv flexibility, while csv columns overriding the defaults ones from DOE data. This requires modifying the `demand_flexibility.cost_up/dn` after reading `flex_cost_up/dn.csv` into the `demand_flexibility` struct, as the total number of flexible bus will be different than the number of columns in the csvs when DOE data is used, and the dimension of the DF need to be modified accordingly. There might be cleaner ways but for this version I just want the logic to be reviewed.

Thanks! I'll be sure to check it out.

> Also the test shows
> 
> > ERROR: Could not find either `git-tree-sha1` or `path` for package DelimitedFiles
> 
> idk if its my problem or something with the pip equivalent for Julia, please let me know if you have seen this before.

The lint test is indicating that `model.jl` and `read.jl` need to be formatted. Don't worry about the Julia nightly test for the time being as it is not an issue on your end. The Julia nightly test is not a required test for the PR to be merged.

src/read.jl
lanesmith: Okay, thank you for looking into that. Perhaps have the comment inside the conditional check (i.e., after `if`, `elseif`, or `else`)? It looks like this particular instance is now gone, but there are a couple other such instances in `read.jl`.

src/model.jl
```diff
@@ -66,21 +66,68 @@ function _make_bus_demand_flexibility_amount(
...
+    # reformat flexibility cost up/dn to appropriate dimensions
+    new_demand_flex_cost_up = zeros(size(demand_flexibility.cost_up, 1), sets.num_bus)
+    new_demand_flex_cost_dn = zeros(size(demand_flexibility.cost_dn, 1), sets.num_bus)
+    
+    # DOE flexibility, convert to MW, DOE timestamps already aligned
+    if demand_flexibility.enable_doe_flexibility
+        doe_flex_pct = Matrix(
+            demand_flexibility.doe_flex_amt[start_index:end_index, 2:end]
+        )
+        doe_flex_mw = doe_flex_pct .*= permutedims(bus_demand[sets.doe_flexible_bus_idx, :])
```
lanesmith: This is just intending to multiply the flexibility percentage at each bus for each time step by the demand at each bus for each time step, right? If so, I think we can just do `doe_flex_mw = doe_flex_pct .* permutedims(bus_demand[sets.doe_flexible_bus_idx, :])`. Using the `.*=` instead of the `.*` modifies how `doe_flex_pct` is stored in memory; I believe it would make `doe_flex_mw` and `doe_flex_pct` equal one another, which doesn't seem to be the intent.

src/model.jl
```diff
@@ -58,19 +58,76 @@ Given a Case object and a DemandFlexibility object, build a matrix of demand fle
...
+    simulation_demand_flex_amt_up = simulation_demand_flex_amt_up[:, sets.flexible_bus_idx]
+    simulation_demand_flex_amt_dn = simulation_demand_flex_amt_dn[:, sets.flexible_bus_idx]
+    new_demand_flex_cost_up = new_demand_flex_cost_up[:, sets.flexible_bus_idx]
+    new_demand_flex_cost_dn = new_demand_flex_cost_dn[:, sets.flexible_bus_idx]
+    
+    # construct new cost Dataframe with appropriate dimension
+    demand_flexibility.cost_up = DataFrames.DataFrame(new_demand_flex_cost_up, :auto)
+    demand_flexibility.cost_dn = DataFrames.DataFrame(new_demand_flex_cost_dn, :auto)
+    
+    DataFrames.rename!(demand_flexibility.cost_dn, Symbol.(sets.flexible_bus_idx))
```
lanesmith: Should this be `DataFrames.rename!(demand_flexibility.cost_up, Symbol.(sets.flexible_bus_idx))`?

src/model.jl
```diff
@@ -66,21 +66,68 @@ function _make_bus_demand_flexibility_amount(
...
+    DataFrames.rename!(demand_flexibility.cost_dn, Symbol.(sets.flexible_bus_idx))
+    DataFrames.insertcols!(
+        demand_flexibility.cost_up, 1, :"UTC Time" => demand_flexibility.flex_amt_up[!, "UTC Time"]
+    )
+    DataFrames.insertcols!(
+        demand_flexibility.cost_dn, 1, :"UTC Time" => demand_flexibility.flex_amt_up[!, "UTC Time"]
+    )    
+
```
lanesmith: Maybe we can change this function (`_make_bus_demand_flexibility_amount`) to also output the cost profiles? We could output the pared-down profiles that are later calculated [here](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/model.jl#L551) and [here](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/model.jl#L561), which would match the similarly pared-down profiles that are already returned by this function. Instead of updating the DataFrames directly in the `DemandFlexibility` object, we could just create local variables for use in this function. Doing this would allow us to keep the structs as immutable (which is quite a bit faster than if the structs are mutable, as far as I understand).

src/types.jl
```diff
@@ -42,7 +42,8 @@ Base.@kwdef struct Storage
     sd_table::DataFrames.DataFrame
 end
 
-Base.@kwdef struct DemandFlexibility
+Base.@kwdef mutable struct DemandFlexibility
```
lanesmith: See my above comment about potentially reverting this struct back to being immutable.

src/model.jl
```diff
@@ -66,21 +66,68 @@ function _make_bus_demand_flexibility_amount(
...
 
-    # Demand flexibility up profile
-    simulation_demand_flex_amt_up = Matrix(
+    # combine two profiles with user input overriding DOE flexibility
+    simulation_demand_flex_amt_up = zeros(end_index - start_index + 1, sets.num_bus)
+    simulation_demand_flex_amt_dn = zeros(end_index - start_index + 1, sets.num_bus)
+    
+    # reformat flexibility cost up/dn to appropriate dimensions
+    new_demand_flex_cost_up = zeros(size(demand_flexibility.cost_up, 1), sets.num_bus)
+    new_demand_flex_cost_dn = zeros(size(demand_flexibility.cost_dn, 1), sets.num_bus)
```
lanesmith: I think we need to account for the possibility that `demand_flexibility.cost_up` and `demand_flexibility.cost_dn` are `nothing`. Maybe keep all the cost-related code together and have an `if` statement to check?

src/model.jl
```diff
@@ -160,15 +207,26 @@ function _make_sets(
     demand_flexibility_enabled =
         isa(demand_flexibility, DemandFlexibility) && demand_flexibility.enabled
     if demand_flexibility_enabled
+        # if DOE profile is used, all buses with valid EIA ID are flexible
+        if demand_flexibility.enable_doe_flexibility
+            doe_flexible_bus_idx = sort(intersect(load_bus_idx, findall(case.bus_eiaid .> 0)))
+        end
```
lanesmith: I think you need to handle the case where demand flexibility is enabled but DOE demand flexibility is not. I think it would probably be something like:

```
if demand_flexibility.enable_doe_flexibility
    doe_flexible_bus_idx = sort(intersect(load_bus_idx, findall(case.bus_eiaid .> 0)))
else
    doe_flexible_bus_idx = nothing
end
```

src/model.jl
```diff
@@ -66,21 +66,68 @@ function _make_bus_demand_flexibility_amount(
...
+    DataFrames.insertcols!(
+        demand_flexibility.cost_up, 1, :"UTC Time" => demand_flexibility.flex_amt_up[!, "UTC Time"]
+    )
+    DataFrames.insertcols!(
+        demand_flexibility.cost_dn, 1, :"UTC Time" => demand_flexibility.flex_amt_up[!, "UTC Time"]
+    )    
+    
     return (
         permutedims(simulation_demand_flex_amt_up),
         permutedims(simulation_demand_flex_amt_dn),
```
lanesmith: If we don't end up making the changes I suggested for this function, I believe we'll still need to return the updated `demand_flexibility`.

lanesmith (commented):
Thank you for adding in that new functionality! In the comments below, I suggest a way to keep the `DemandFlexibility` struct as immutable. Please let me know if you have any questions.

lanesmith:
We should probably also add a warning (not an error) in `read.jl`'s `read_demand_flexibility` in the event that DOE flexibility is enabled and demand flexibility profiles are provided. This warning would let the user know the policy for taking the provided demand flexibility profiles over the DOE flexibility profiles.

src/model.jl
dongqi-wu: ah this seems to be a typo..

src/model.jl
dongqi-wu: I am going to fix the mutable problem. but in julia do I need to also return an updated struct/DF after changing it in a function to keep the change?

lanesmith: I believe so? I haven't tested it (though we could probably do a pretty simple test), but I believe the function creates a copy of the `DemandFlexibility` object that is local to the function. If we wanted to keep changes made to the variable, I believe we need to return it from the function

dongqi-wu: hmm that's annoying if so. even the ancient C doesn't do this.

lanesmith: Ah, it appears that you're right: 

http://web.mit.edu/julia_v0.6.2/julia/share/doc/julia/html/en/manual/functions.html#:~:text=Julia%20function%20arguments%20follow%20a,identical%20to%20the%20passed%20values.

Though if the changes are made, this doesn't seem like it will come up.

src/model.jl
```diff
@@ -58,18 +58,48 @@ Given a Case object and a DemandFlexibility object, build a matrix of demand fle
...
+    csv_flex_amt_dn = Matrix(demand_flexibility.flex_amt_dn[start_index:end_index, 2:end])
+
+    for i in 1:length(sets.csv_flexible_bus_idx)
+        simulation_demand_flex_amt_up[:, sets.csv_flexible_bus_idx[i]] = csv_flex_amt_up[
+            :, i
+        ]
+        simulation_demand_flex_amt_dn[:, sets.csv_flexible_bus_idx[i]] = csv_flex_amt_dn[
+            :, i
+        ]
+    end
```
lanesmith: Perhaps I'm missing something, but with the inclusion of the ability to use the DOE demand flexibility data, it is no longer guaranteed that `demand_flexibility.flex_amt_up` and `demand_flexibility.flex_amt_dn` will not be nothing if demand flexibility is enabled. Do we need to add a check here to see if `demand_flexibility.flex_amt_up` and `demand_flexibility.flex_amt_dn` are not nothing? Otherwise, it seems that there would be an error because `demand_flexibility.flex_amt_up` and `demand_flexibility.flex_amt_dn` would not have indices.

dongqi-wu: sorry for waiting so long, it seems to be a minor problem though. there is also another typo that I did not know until I do a test-run for this issue. if the logics seems OK I will fix the conflicts and rebase tmr.

lanesmith:
@dongqi-wu, thank you for your latest commit! For the most part, I think this is good to go on my end. Please be sure to address my unresolved comment about formatting and my comment about adding a warning above. Additionally, I'm not sure what happened, but it seems that you already added in some of the latest changes from the develop branch (at least from what I can tell by the latest commit's changes to `read.jl`). Perhaps this will be resolved when you rebase.

Once the above is addressed and the code is rebased, this should be able to be approved on my end.

dongqi-wu:
> @dongqi-wu, thank you for your latest commit! For the most part, I think this is good to go on my end. Please be sure to address my unresolved comment about formatting and my comment about adding a warning above. Additionally, I'm not sure what happened, but it seems that you already added in some of the latest changes from the develop branch (at least from what I can tell by the latest commit's changes to `read.jl`). Perhaps this will be resolved when you rebase.
> 
> Once the above is addressed and the code is rebased, this should be able to be approved on my end.

Sure I will check the comments again and run juliaformatter before trying to rebase. Regarding the latest changes, I tried to copy them into read.jl before the last commit to see if the conflict still appears.

src/read.jl
```diff
@@ -10,50 +10,45 @@ function read_case(filepath)
...
     # Buses
-    case["busid"] = convert(Array{Int,1}, mpc["bus"][:, 1])
-    case["bus_demand"] = mpc["bus"][:, 3]
-    case["bus_zone"] = convert(Array{Int,1}, mpc["bus"][:, 7])
+    bus = CSV.File(joinpath(filepath, "bus.csv"))
+    case["busid"] = convert(Array{Int,1}, bus.bus_id)
+    case["bus_demand"] = convert(Array{Float64,1}, bus.Pd)
+    case["bus_zone"] = convert(Array{Int,1}, bus.zone_id)
+    try
+        case["bus_eiaid"] = convert(Array{Int,1}, mpc["bus"][:, 18])
```
lanesmith: Instead of using an index number (i.e., 18), use the name of the column. So this line would be something like `case["bus_eiaid"] = convert(Array{Int,1}, bus.bus_eiaid)` (replace "bus_eiaid" with the correct column name if it is different)

dongqi-wu: > Instead of using an index number (i.e., 18), use the name of the column. So this line would be something like `case["bus_eiaid"] = convert(Array{Int,1}, bus.bus_eiaid)` (replace "bus_eiaid" with the correct column name if it is different)

I have changed the column name and rebased just now. Please check if the re-basing is working properly. I will start working on PowerSimData while this is underway.

lanesmith (commented):
This looks good to me. @BainanXia, I'm not sure if you wanted to look this over as well, but once you give the go-ahead, I'm happy to approve this one my end.

@dongqi-wu, I also want to check that you've been able to test this, especially since you rebased to our latest version of develop.

src/model.jl
```diff
@@ -58,18 +58,54 @@ Given a Case object and a DemandFlexibility object, build a matrix of demand fle
...
+                :, i
+            ]
+            simulation_demand_flex_amt_dn[:, sets.doe_flexible_bus_idx[i]] = doe_flex_mw[
+                :, i
+            ]
+        end
+    end
+
+    # replace columns with user input numbers if applicable
+    if !isnothing(demand_flexibility.flex_amt_up)
```
BainanXia, line 88: @lanesmith We are assuming both `flex_amt_up` and `flex_amt_dn` should be specified simultaneously by the user  here, i.e. there shouldn't be the case that `demand_flexibility.flex_amt_up == nothing` but `demand_flexibility.flex_amt_dn` is specified, right?

src/read.jl
```diff
@@ -26,6 +26,11 @@ function read_case(filepath)
     case["busid"] = convert(Array{Int,1}, bus.bus_id)
     case["bus_demand"] = convert(Array{Float64,1}, bus.Pd)
     case["bus_zone"] = convert(Array{Int,1}, bus.zone_id)
+    try
+        case["bus_eiaid"] = convert(Array{Int,1}, bus.eia_id)
+    catch e
+        case["bus_eiaid"] = zeros(size(case["busid"], 1), 1)
+    end
```
BainanXia, line 33: Just to clarify the logic here: we are assuming either get EIA id for all buses from `PowerSimData` (precomputed ones + new buses assigned by nearest neighbor logic during scenario creation) or EIA id is not defined at all.

src/read.jl
```diff
@@ -197,10 +237,12 @@ function read_demand_flexibility(filepath, interval)::DemandFlexibility
     end
 
     # If demand flexibility is enabled but at least one demand flexibility profile is nothing
-    if demand_flexibility["enabled"] == true && (
-        isnothing(demand_flexibility["flex_amt_up"]) ||
-        (isnothing(demand_flexibility["flex_amt_dn"]))
-    )
+    if demand_flexibility["enabled"] == true &&
```
BainanXia, line 221: Not sure whether we should move this entire `if` block up before l.170: "Try loading DOE flexibility profile if enabled", since we set `demand_flexibility["enabled"]` here.

src/read.jl
```diff
@@ -426,42 +476,101 @@ function reformat_demand_flexibility_input(
...
                     # identify flexible bus by their total cost and append to list
                     if any(x -> x > 0, converted_zone_cost[:, i])
                         append!(flex_bus_idx, i)
                     end
                 end
-                eq_bus_df = DataFrames.DataFrame(converted_zone_cost[:, flex_bus_idx])
+
+                # if DOE flexibility is used, add columns for flexible buses there
+
+                # find all flexible buses
```
BainanXia, line 512: This is minor and we might refactor in future PRs: it seems the logic of finding all flexible buses is used several times throughout the module, not sure we should factor it out to reduce code duplication.

src/read.jl
```diff
@@ -426,42 +476,101 @@ function reformat_demand_flexibility_input(
...
+    if demand_flexibility.enable_doe_flexibility == true
+        demand_flex_field = getfield(demand_flexibility, Symbol("doe_flex_amt"))
+
+        # all load buses with valid EIA ID 
+        flexible_bus_idx = intersect(sets.load_bus_idx, findall(case.bus_eiaid .> 0))
+        flexible_bus_num = length(flexible_bus_idx)
+
+        # bus flexibility Percentage matrix
+        doe_bus_flexibility = zeros(size(case.demand, 1), flexible_bus_num)
+        doe_eiaid = [parse(Int, x) for x in names(demand_flex_field)[2:end]]
```
BainanXia: Maybe I missed something, it seems `doe_eiaid` is not used elsewhere.

BainanXia (commented):
I went through the code and it looks nice and clean, thanks for the effort. I left a few comments simply to clarify the logic which should be minor. @lanesmith Feel free to press the button if there is no further concerns. I also checked the feature branch is successfully rebased onto the latest `develop`.

rouille:
@jenhagg, we will need to include an extra column for the `bus` data frame when we export the grid.pkl to CSVs. Just a reminder that we need to work on it

src/model.jl
lanesmith: At least for this iteration, yes. There should be some logic in `read.jl` that throws an error if `flex_amt_up` and `flex_amt_dn` aren't both specified. In future iterations, we may want different logic so that load shedding can be considered. Perhaps that would be a different input though.

lanesmith:
> @jenhagg, we will need to include an extra column for the `bus` data frame when we export the grid.pkl to CSVs. Just a reminder that we need to work on it

It's been a while, but I just want to link [PowerSimData PR #623](https://github.com/Breakthrough-Energy/PowerSimData/pull/623), which specifies this feature, albeit under the old convention.

lanesmith (approved):
Looks good to me. Once @BainanXia's comments are addressed, this should be good to go. Please also make sure that any tests you have are able to work now that your branch is rebased.

dongqi-wu:
> Looks good to me. Once @BainanXia's comments are addressed, this should be good to go. Please also make sure that any tests you have are able to work now that your branch is rebased.

sure I will fix them soon.

src/read.jl
dongqi-wu: > Just to clarify the logic here: we are assuming either get EIA id for all buses from `PowerSimData` (precomputed ones + new buses assigned by nearest neighbor logic during scenario creation) or EIA id is not defined at all.

if the csv does not have eiaid then they all get assigned with 0 which means none of them belong to any LSE and thus the DOE flexibility will not be used.

dongqi-wu:
> Looks good to me. Once @BainanXia's comments are addressed, this should be good to go. Please also make sure that any tests you have are able to work now that your branch is rebased.

image
for the testing, I just found the refactored code give the error below using the old input files. was the file changed in the past?

>     # Generator costs
>    gencost = DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))
>    case["gencost"] = convert(Matrix{Float64}, gencost)

lanesmith:
> image for the testing, I just found the refactored code give the error below using the old input files. was the file changed in the past?
> 
> > ```
> > # Generator costs
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gencost = DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))
> > case["gencost"] = convert(Matrix{Float64}, gencost)

Yeah, the change to `REISE.jl` that you rebased changed how we consider input files. Now, instead of using the `case.mat` files, we output grid information from `PowerSimData` as a .pkl file. If you use the Python wrapper, the .pkl is taken as input and is used to produce the different .csv files that you now see read in in `read.jl`. If you use the Julia interface, you need to have those .csv files. We can get you new test files for the the Texas grid, though you'll need to add in the column of EIA flexibility information to the appropriate .csv file.

dongqi-wu:
> > image for the testing, I just found the refactored code give the error below using the old input files. was the file changed in the past?
> > > ```
> > > # Generator costs
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > gencost = DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))
> > > case["gencost"] = convert(Matrix{Float64}, gencost)
> 
> Yeah, the change to `REISE.jl` that you rebased changed how we consider input files. Now, instead of using the `case.mat` files, we output grid information from `PowerSimData` as a .pkl file. If you use the Python wrapper, the .pkl is taken as input and is used to produce the different .csv files that you now see read in in `read.jl`. If you use the Julia interface, you need to have those .csv files. We can get you new test files for the the Texas grid, though you'll need to add in the column of EIA flexibility information to the appropriate .csv file.

Thank you. How do I get the new files? I can run my local script to get the EIA ID column myself.

lanesmith:
> Thank you. How do I get the new files? I can run my local script to get the EIA ID column myself.

Sorry I didn't share it yesterday. I thought you said you had the .csv files. Using `PowerSimData`, you can create the .pkl file as follows:

```
import os
import pickle
import powersimdata
from powersimdata import Scenario

# Create grid object
scenario = Scenario()
scenario.set_grid(grid_model="usa_tamu", interconnect="Texas")
grid = scenario.get_grid()

# Dump in a .pkl file
filepath = INSERT FILE PATH STRING HERE
with open(os.path.join(filepath, "grid.pkl"), "wb") as f:
    pickle.dump(grid, f, protocol=pickle.HIGHEST_PROTOCOL)
```

Once you have the .pkl file, you should be able to run `REISE.jl` using the Python interface, as is specified [here](https://breakthrough-energy.github.io/docs/reisejl/index.html#id3). If you just want to run `REISE.jl` using the Julia interface, you will need to run the [appropriate functions](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/call.py#L32-#L77) in `pyreisejl/utility/call.py` to create the .csv files from the .pkl file. This latter path might be preferred so you can add the EIA ID column.

dongqi-wu:
> > Thank you. How do I get the new files? I can run my local script to get the EIA ID column myself.
> 
> Sorry I didn't share it yesterday. I thought you said you had the .csv files. Using `PowerSimData`, you can create the .pkl file as follows:
> 
> ```
> import os
> import pickle
> import powersimdata
> from powersimdata import Scenario
> 
> # Create grid object
> scenario = Scenario()
> scenario.set_grid(grid_model="usa_tamu", interconnect="Texas")
> grid = scenario.get_grid()
> 
> # Dump in a .pkl file
> filepath = INSERT FILE PATH STRING HERE
> with open(os.path.join(filepath, "grid.pkl"), "wb") as f:
>     pickle.dump(grid, f, protocol=pickle.HIGHEST_PROTOCOL)
> ```
> 
> Once you have the .pkl file, you should be able to run `REISE.jl` using the Python interface, as is specified [here](https://breakthrough-energy.github.io/docs/reisejl/index.html#id3). If you just want to run `REISE.jl` using the Julia interface, you will need to run the [appropriate functions](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/call.py#L32-#L77) in `pyreisejl/utility/call.py` to create the .csv files from the .pkl file. This latter path might be preferred so you can add the EIA ID column.

Thanks!

dongqi-wu:
> > Thank you. How do I get the new files? I can run my local script to get the EIA ID column myself.
> 
> Sorry I didn't share it yesterday. I thought you said you had the .csv files. Using `PowerSimData`, you can create the .pkl file as follows:
> 
> ```
> import os
> import pickle
> import powersimdata
> from powersimdata import Scenario
> 
> # Create grid object
> scenario = Scenario()
> scenario.set_grid(grid_model="usa_tamu", interconnect="Texas")
> grid = scenario.get_grid()
> 
> # Dump in a .pkl file
> filepath = INSERT FILE PATH STRING HERE
> with open(os.path.join(filepath, "grid.pkl"), "wb") as f:
>     pickle.dump(grid, f, protocol=pickle.HIGHEST_PROTOCOL)
> ```
> 
> Once you have the .pkl file, you should be able to run `REISE.jl` using the Python interface, as is specified [here](https://breakthrough-energy.github.io/docs/reisejl/index.html#id3). If you just want to run `REISE.jl` using the Julia interface, you will need to run the [appropriate functions](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/call.py#L32-#L77) in `pyreisejl/utility/call.py` to create the .csv files from the .pkl file. This latter path might be preferred so you can add the EIA ID column.

@lanesmith any idea how the gencost functions in the previous update works?
image
the Python function only produces one `gencost.csv` from the pickle file, but it seems the `read.jl` is taking two at a time now.

edit: the previous read gencost.csv line gave an error as well about converting Dataframe to Matrix. The current version is after rebasing on top of the newest commit, but the code is not tested due to the input file mismatch.

lanesmith:
> @lanesmith any idea how the gencost functions in the previous update works? image the Python function only produces one `gencost.csv` from the pickle file, but it seems the `read.jl` is taking two at a time now.
> 
> edit: the previous read gencost.csv line gave an error as well about converting Dataframe to Matrix. The current version is after rebasing on top of the newest commit, but the code is not tested due to the input file mismatch.

Hi @dongqi-wu, sorry I'm just now getting back to you. We recently updated `REISE.jl` and `PowerSimData` so that some of the processes we had to linearize gencosts and update pmins and ramp rates were moved from `REISE.jl` to `PowerSimData`. You can now run the following to create the necessary .pkl file (i.e., will include appropriate `grid.gencost["after"]` and updated `pmin` and `ramp_30` values that will reflect what `REISE.jl` used to implement):

```python
import os
import pickle
import powersimdata
from powersimdata import Scenario
from powersimdata.input.configure import linearize_gencost, adjust_pmin, adjust_ramp30

# Create grid object
scenario = Scenario()
scenario.set_grid(grid_model="usa_tamu", interconnect="Texas")
grid = scenario.get_grid()
adjust_pmin(grid)
adjust_ramp30(grid.plant)
grid.gencost["after"] = linearize_gencost(grid.gencost["before"], grid.plant, num_segments=1)

# Dump in a .pkl file
filepath = INSERT FILE PATH STRING HERE
with open(os.path.join(filepath, "grid.pkl"), "wb") as f:
    pickle.dump(grid, f, protocol=pickle.HIGHEST_PROTOCOL)
```

Be sure that you've got the latest version of `PowerSimData`'s develop branch before you do this. Once you have the new .pkl file, you should be able to run `REISE.jl`'s Python interface as specified before. This should take care of creating those different gencost .csv files (as well as the other .csv files) you were asking about.

dongqi-wu:
> > @lanesmith any idea how the gencost functions in the previous update works? image the Python function only produces one `gencost.csv` from the pickle file, but it seems the `read.jl` is taking two at a time now.
> > edit: the previous read gencost.csv line gave an error as well about converting Dataframe to Matrix. The current version is after rebasing on top of the newest commit, but the code is not tested due to the input file mismatch.
> 
> Hi @dongqi-wu, sorry I'm just now getting back to you. We recently updated `REISE.jl` and `PowerSimData` so that some of the processes we had to linearize gencosts and update pmins and ramp rates were moved from `REISE.jl` to `PowerSimData`. You can now run the following to create the necessary .pkl file (i.e., will include appropriate `grid.gencost["after"]` and updated `pmin` and `ramp_30` values that will reflect what `REISE.jl` used to implement):
> 
> ```python
> import os
> import pickle
> import powersimdata
> from powersimdata import Scenario
> from powersimdata.input.configure import linearize_gencost, adjust_pmin, adjust_ramp30
> 
> # Create grid object
> scenario = Scenario()
> scenario.set_grid(grid_model="usa_tamu", interconnect="Texas")
> grid = scenario.get_grid()
> adjust_pmin(grid)
> adjust_ramp30(grid.plant)
> grid.gencost["after"] = linearize_gencost(grid.gencost["before"], grid.plant, num_segments=1)
> 
> # Dump in a .pkl file
> filepath = INSERT FILE PATH STRING HERE
> with open(os.path.join(filepath, "grid.pkl"), "wb") as f:
>     pickle.dump(grid, f, protocol=pickle.HIGHEST_PROTOCOL)
> ```
> 
> Be sure that you've got the latest version of `PowerSimData`'s develop branch before you do this. Once you have the new .pkl file, you should be able to run `REISE.jl`'s Python interface as specified before. This should take care of creating those different gencost .csv files (as well as the other .csv files) you were asking about.

Thanks, I will try and get back to you.

edit:
I have tested the new input files and they all work (with or without DOE/CSV data). Please see the newest commit, it should also pass the style check as well.

lanesmith:
> Thanks, I will try and get back to you.
> 
> edit: I have tested the new input files and they all work (with or without DOE/CSV data). Please see the newest commit, it should also pass the style check as well.

Thanks for verifying that, @dongqi-wu. I've looked over the latest commit and it looks good to me. So long as the checks pass, I think this is good to merge. @BainanXia, do you have any further feedback? Also, @BainanXia, I am not able to run the above checks, can you allow them to run?

BainanXia:
> > Thanks, I will try and get back to you.
> > edit: I have tested the new input files and they all work (with or without DOE/CSV data). Please see the newest commit, it should also pass the style check as well.
> 
> Thanks for verifying that, @dongqi-wu. I've looked over the latest commit and it looks good to me. So long as the checks pass, I think this is good to merge. @BainanXia, do you have any further feedback? Also, @BainanXia, I am not able to run the above checks, can you allow them to run?

It looks good to me as well and I can't run the checks either. It says "Waiting for status to be reported". Not sure what's going on here. Maybe @jenhagg has a better idea?

jenhagg:
> > > Thanks, I will try and get back to you.
> > > edit: I have tested the new input files and they all work (with or without DOE/CSV data). Please see the newest commit, it should also pass the style check as well.
> > 
> > 
> > Thanks for verifying that, @dongqi-wu. I've looked over the latest commit and it looks good to me. So long as the checks pass, I think this is good to merge. @BainanXia, do you have any further feedback? Also, @BainanXia, I am not able to run the above checks, can you allow them to run?
> 
> It looks good to me as well and I can't run the checks either. It says "Waiting for status to be reported". Not sure what's going on here. Maybe @jenhagg has a better idea?

Not sure why they wouldn't be triggered other than github being unreliable. Two options I can think of are pushing a new commit, or if that doesn't work, adding a `workflow_dispatch` trigger to the checks so we can run them manually.

BainanXia (approved):
Thanks everyone for the effort and patience 🚀

PR #183, opened 2022-08-25
Update python version for testing and dependencies in requirements file

rouille:

### Purpose
Update python versions for testing. More specifically, remove 3.7 and add 3.9 and 3.10. The `h5py` and `pytest` packages in the requirements file had to be updated to get the checks running.  

### What the code is doing
N/A

### Testing
Existing tests

### Where to look
The requirements file.:
* The latest `pytest` will be installed
* The latest version of `h5py`, greater than or equal to 3.7., but still in the 3 version will be installed

### Usage Example/Visuals
N/A

### Time estimate
5min

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          version: [3.7, 3.8]
+          version: [3.8, 3.9, '3.10']
     steps:
       - uses: actions/checkout@v2
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-h5py~=2.10.0
+h5py~=3.7
 ipython~=7.0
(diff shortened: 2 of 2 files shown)


lanesmith (approved):
Looks good to me!

lanesmith:
Does this mean we are dropping support for Python 3.7? If so, we might want to update the Python installation page in the docs: https://breakthrough-energy.github.io/docs/user/python_guide.html

rouille:
> Does this mean we are dropping support for Python 3.7? If so, we might want to update the Python installation page in the docs: https://breakthrough-energy.github.io/docs/user/python_guide.html

Good point. We stop supporting 3.7 to follow what we do in other public packages.

jenhagg (approved):

jenhagg:
I'm not an admin on this repo but looks like we need to update the required checks to remove 3.7, and possibly add 3.10.

PR #185, opened 2022-09-29
Replace case.mat with grid.pkl

jenhagg:

### Purpose
Remove dependence on column indices when reading case.mat by converting a grid.pkl to csv files with the appropriate columns, and accessing columns by name.

### What the code is doing
* Convert grid.pkl to csv files and read those in read.jl
* Keep the case.mat which is still used by `save_input_mat`

### Testing
Ran a scenario in a container

### Time estimate
30 mins

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         version:
           - '1.5'
-          - 'nightly'
+          # - 'nightly'
         os:
           - ubuntu-latest
     steps:
--- a/pyreisejl/utility/call.py
+++ b/pyreisejl/utility/call.py
@@ -23,9 +23,56 @@ def _record_scenario(scenario_id, runtime):
     # Update status in ExecuteList.csv on server
     insert_in_file(const.EXECUTE_LIST, scenario_id, "status", "finished")
 
-    hours, minutes, seconds = sec2hms(runtime)
+    hours, minutes, _ = sec2hms(runtime)
     insert_in_file(
         const.SCENARIO_LIST, scenario_id, "runtime", "%d:%02d" % (hours, minutes)
+    )
+
+
+cols = {
+    "branch": ["branch_id", "from_bus_id", "to_bus_id", "x", "rateA"],
+    "dcline": ["dcline_id", "from_bus_id", "to_bus_id", "Pmin", "Pmax"],
+    "bus": ["bus_id", "Pd", "zone_id"],
+    "plant": [
+        "plant_id",
+        "bus_id",
+        "status",
+        "Pmin",
+        "Pmax",
+        "type",
+        "ramp_30",
+        "GenFuelCost",
+        "GenIOB",
+        "GenIOC",
+        "GenIOD",
+    ],
+}
+
+
+def _save(path, name, df):
+    df = df.reset_index()
+    df = df.loc[:, cols.get(name, df.columns)]
+    df.to_csv(os.path.join(path, f"{name}.csv"), index=False)
+
+
+def pkl_to_csv(path):
+    with open(os.path.join(path, "grid.pkl"), "rb") as f:
+        grid = pickle.load(f)
+    _save(path, "branch", grid.branch)
+    _save(path, "dcline", grid.dcline)
+    _save(path, "bus", grid.bus)
+    _save(path, "plant", grid.plant)
+    _save(path, "gencost", grid.gencost["before"])
+
+
+def pkl_to_case_mat(path):
+    with open(os.path.join(path, "grid.pkl"), "rb") as f:
+        grid = pickle.load(f)
+
+    _, _ = create_case_mat(
+        grid,
+        filepath=os.path.join(path, "case.mat"),
+        storage_filepath=os.path.join(path, "case_storage.mat"),
     )
 
 
@@ -41,15 +88,8 @@ def main(args):
         args.input_dir = scenario_args[3]
         args.execute_dir = scenario_args[4]
 
-        # Create grid data for simulatio
-        with open(os.path.join(args.input_dir, "grid.pkl"), "rb") as f:
-            grid = pickle.load(f)
-
-        _, _ = create_case_mat(
-            grid,
-            filepath=os.path.join(args.input_dir, "case.mat"),
-            storage_filepath=os.path.join(args.input_dir, "case_storage.mat"),
-        )
+        pkl_to_csv(args.input_dir)
+        pkl_to_case_mat(args.input_dir)
 
         # Update status in ExecuteList.csv on server
         insert_in_file(const.EXECUTE_LIST, args.scenario_id, "status", "running")
--- a/pyreisejl/utility/call.py
+++ b/pyreisejl/utility/call.py
@@ -58,6 +58,10 @@ def _save(path, name, df):
     df.to_csv(os.path.join(path, f"{name}.csv"), index=False)
 
 
+def _save_storage(path, name, df):
+    df.to_csv(os.path.join(path, f"{name}.csv"), index=False)
+
+
 def pkl_to_csv(path):
     with open(os.path.join(path, "grid.pkl"), "rb") as f:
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         version:
           - '1.5'
-          - 'nightly'
+          # - 'nightly'
```
rouille, line 19: Thanks, that was annoying

pyreisejl/utility/call.py
```diff
@@ -21,15 +23,66 @@ def _record_scenario(scenario_id, runtime):
...
+        "type",
+        "ramp_30",
+        "GenFuelCost",
+        "GenIOB",
+        "GenIOC",
+        "GenIOD",
+    ],
+}
+
+drop_cols = {"gencost": ["plant_id", "interconnect"]}
```
rouille, line 51: Why do we drop columns for `gencost` instead of selecting columns in `cols` as we do for the other data frames?

BainanXia: I guess it's shorter, i.e. between a white list in selecting columns vs. a black list in dropping columns, always go for the shorter one?

src/read.jl
```diff
@@ -2,58 +2,45 @@
...
 
     # AC branches
-    # dropdims() will remove extraneous dimension
-    case["branchid"] = dropdims(mpc["branchid"]; dims=2)
-    # convert() will convert float array to int array
-    case["branch_from"] = convert(Array{Int,1}, mpc["branch"][:, 1])
-    case["branch_to"] = convert(Array{Int,1}, mpc["branch"][:, 2])
-    case["branch_reactance"] = mpc["branch"][:, 4]
-    case["branch_rating"] = mpc["branch"][:, 6]
+    branch = CSV.File(joinpath(filepath, "branch.csv"))
```
BainanXia, line 9: Since we use `PyCall` anyways in `REISE.jl`, not sure whether we should look into reading pkl files directly in julia via `PyCall`.

.github/workflows/test.yml
lanesmith: Agreed! Thanks for doing this. Since 1.6.7 is the current long-term support release and 1.8.2 is the current stable release, maybe we should add Julia tests for versions 1.6, 1.7, and 1.8? We should probably also think about bumping the Julia version from 1.5 to 1.6 at some point, so adding the test would be a helpful first check.

src/read.jl
rouille: Interesting. It seems to be well supported.

pyreisejl/utility/call.py
rouille: I see. Then you have the issue of considering columns that are not needed if we augment the `gencost` data frame in the future for some reason.

BainanXia: > I see. Then you have the issue of considering columns that are not needed if we augment the `gencost` data frame in the future for some reason.

Indeed. But it is unlikely the case unless we would like to support different cost curve types for different entries.

src/read.jl
lanesmith: The only problem I could potentially see with this would be in the stand-alone version of `REISE.jl`. If we had it so that only a .pkl file was used, Python would need to be used for all functionality of `REISE.jl`, not just the Python wrappers. I don't think this is a problem (when I use stand-alone `REISE.jl`, I use the Python wrappers), but we would probably need to change the documentation.

src/read.jl
```diff
@@ -2,58 +2,45 @@
...
+    case["gen_bus"] = convert(Array{Int,1}, plant.bus_id)
+    case["gen_status"] = convert(BitArray{1}, plant.status)
+    case["gen_pmax"] = convert(Array{Float64,1}, plant.Pmax)
+    case["gen_pmin"] = convert(Array{Float64,1}, plant.Pmin)
+    case["gen_ramp30"] = convert(Array{Float64,1}, plant.ramp_30)
 
     # Generator costs
-    case["gencost"] = mpc["gencost"]
+    gencost = CSV.File(joinpath(filepath, "gencost.csv"))
+    df = DataFrames.DataFrame(gencost)
```
lanesmith: This can be done in one line: `DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))`. 

EDIT: If you do this, be sure to change the `df` in the following's line's creation of `case["gencost"]`.

lanesmith:
Does `pickle` need to be added to the `requirements.txt` file?

rouille:
> Does `pickle` need to be added to the `requirements.txt` file?

No, it is part of the standard library.

lanesmith:
> No, it is part of the standard library.

Ah, okay. Thank you for clarifying!

pyreisejl/utility/call.py
```diff
@@ -38,6 +91,9 @@ def main(args):
         args.input_dir = scenario_args[3]
         args.execute_dir = scenario_args[4]
 
+        pkl_to_csv(args.input_dir)
+        pkl_to_case_mat(args.input_dir)
```
lanesmith, line 104: Is this included for `save_input_mat`?

jenhagg: Yep - I think that's the only thing that still uses case.mat.

pyreisejl/utility/call.py
jenhagg: Yeah, only reason is that was the first thing I thought of and it's probably shorter to drop cols from gencost compared with the other ones. It also kinda made sense when just looking at types.jl where we assign the whole gencost matrix directly, compared to the other fields which are assigned by column name.

src/read.jl
jenhagg: I think it's simpler to use csv as the common interface rather than something at the language level like PyCall. Not that it wouldn't necessarily work well, but I feel like relying only on a standard data format just rules out a potential class of issues that are inherent with that. We could always reevaluate if the explicit mapping gets to be too much maintenance..

pyreisejl/utility/call.py
lanesmith: Great, thanks! I believe we discussed it a week ago, but can you remind me why we still need to generate a `case.mat` file at all?

jenhagg: Yeah we use it to create the `input.mat` which is converted to the `grid.mat` in `extract_data.py`. So eventually we will replace that too but I kept it for now to keep the PR a more reasonable size.

lanesmith: Sounds good! Thank you for clarifying!

jenhagg:
I added some changes to handle storage, and some minor fixes. Lemme know what ya'll think.

.github/workflows/test.yml
rouille: Sounds good. Let's tackle that in a separate PR. Do you want to take care of it @lanesmith?

src/read.jl
```diff
@@ -71,26 +58,24 @@ function read_case(filepath)
...
-        case_storage_file = MAT.matopen(joinpath(filepath, "case_storage.mat"))
-        storage_mat_data = read(case_storage_file, "storage")
-        println("...loading case_storage.mat")
+        println("...loading storage")
+        gen = DataFrames.DataFrame(CSV.File(joinpath(filepath, "storage_gen.csv")))
+
         # Convert N x 1 array of strings into 1D array of Symbols (length N)
-        column_symbols = Symbol.(vec(storage_mat_data["sd_table"]["colnames"]))
-        data = convert(Array{Float64,2}, storage_mat_data["sd_table"]["data"])
+        sd_path = joinpath(filepath, "StorageData.csv")
```
rouille: Do we need to declare a new variable here?

rouille (approved):
Looks good. Thanks.

src/read.jl
jenhagg: Nope - I was using it twice in a previous version but now it can be one line

BainanXia (approved):
Thanks

PR #187, opened 2022-11-03
Remove grid modifications and refactor extract_data

jenhagg:

### Purpose
Simplify the file management by using a grid.pkl as the engine input, removing the modifications to cost curves and pmin values by requiring those already be done, and removing the intermediate step of creating an input.mat which then has to be converted to a grid.mat and read back into a grid object by PowerSimData. 

### What the code is doing
* Deleted unused code
* In extract_data, use a grid object to generate timestamp indices for results
* change the type of gencost arrays in types.jl to a julia DataFrame and replace positional indices in model.jl with column names
* remove execute-dir and matlab-dir command line arguments, still having defaults that allow a user to run a simulation provided they have a grid.pkl and profiles in csv format
* misc refactoring of the way we handle standalone vs scenario based usage

### Testing
Ran a simulation in a container using both PowerSimData and directly, using call.py and extract_data.py. I used the [demo_texas.py](https://github.com/Breakthrough-Energy/plug/blob/main/standalone/demo_texas.py) script as usual, and the following commands for standalone (after creating a directory with the input files)

```
root@reisejl:/app# python pyreisejl/utility/call.py -s '2016-01-01' -e '2016-01-01 03:00:00' -int 1 -i '/app/test2' --solver clp
root@reisejl:/app# python pyreisejl/utility/extract_data.py -s '2016-01-01' -e '2016-01-01 03:00:00' -i '/app/test2' -o '/app/test2_output'
```

### Time estimate
~1 hour.


docs/usage.rst
```diff
@@ -95,8 +93,8 @@ The full list of arguments can be accessed via :bash:`pyreisejl/utility/call.py
 .. code-block:: text
 
    usage: call.py [-h] [-s START_DATE] [-e END_DATE] [-int INTERVAL] [-i INPUT_DIR]
-   [-x EXECUTE_DIR] [-t THREADS] [-d] [-o OUTPUT_DIR] [-m MATLAB_DIR] [-k]
-   [--solver SOLVER] [-j JULIA_ENV] [-l LINEARIZATION_SEGMENTS]
+   [-t THREADS] [-d] [-o [OUTPUT_DIR]] [-k]
```
rouille: Why do we need brackets around `OUTPUT_DIR`?

docs/usage.rst
```diff
@@ -135,15 +129,10 @@ The full list of arguments can be accessed via :bash:`pyreisejl/utility/call.py
                             will be deleted. The extraction process can be memory
                             intensive. This is optional and defaults to False if the
                             flag is omitted.
-      -o OUTPUT_DIR, --output-dir OUTPUT_DIR
+      -o [OUTPUT_DIR], --output-dir [OUTPUT_DIR]
```
rouille: same question about the bracket

pyreisejl/utility/call.py
```diff
@@ -29,122 +24,60 @@ def _record_scenario(scenario_id, runtime):
...
-        grid = pickle.load(f)
-
-    _, _ = create_case_mat(
-        grid,
-        filepath=os.path.join(path, "case.mat"),
-        storage_filepath=os.path.join(path, "case_storage.mat"),
-    )
+def _ensure_required_args(args):
+    """Check to make sure all necessary arguments are there:
+        (start_date, end_date, interval, input_dir)
```
rouille: No indent

rouille:
Do we keep the transfer of the overwriting of the ramp rate into PowerSimData for a separate PR?

pyreisejl/utility/converters.py
```diff
@@ -1,132 +1,51 @@
...
+
+def pkl_to_csv(path):
+    with open(os.path.join(path, "grid.pkl"), "rb") as f:
+        grid = pickle.load(f)
+    _save(path, "branch", grid.branch)
+    _save(path, "dcline", grid.dcline)
+    _save(path, "bus", grid.bus)
+    _save(path, "plant", grid.plant)
+    _save(path, "gencost_orig", grid.gencost["before"])
+    _save(path, "gencost", grid.gencost["after"])
```
rouille: This is out of the scope of this PR but I would keep `before` and `after` for `gencost` as we have in the `Grid` object. Also I think it is more explicit.

src/prepare.jl
```diff
@@ -3,23 +3,9 @@
...
-    case["gen_pmin"][geo_idx] = 0.95 * (case["gen_pmax"][geo_idx])
-    offstatus_idx = case["gen_status"] .== 0
-    case["gen_pmin"][offstatus_idx] .= 0
-
-    # Save original gencost to gencost_orig
-    case["gencost_orig"] = copy(case["gencost"])
-    # Modify gencost based on desired linearization structure
-    case["gencost"] = _linearize_gencost(case; num_segments=num_segments)
-
     # Relax ramp constraints
```
rouille: Should be moved into PowerSimData or apply to the data fetch from Zenodo

src/model.jl
```diff
@@ -85,13 +85,12 @@ Given a Case object, an index of segments (e.g. 1:3), and a vector of widths
...
-    COST = 5        # Positional index from mpc.gencost
     segment_slope = zeros(length(case.genid), length(segment_idx))
     for i in segment_idx
         segment_slope[:, i] = (
-            (2 * case.gencost_orig[:, COST] .* case.gen_pmin) +
-            case.gencost_orig[:, COST + 1] +
-            (2 * i - 1) * case.gencost_orig[:, COST] .* segment_width
+            (2 * case.gencost_orig.c2 .* case.gen_pmin) +
+            case.gencost_orig.c1 +
+            (2 * i - 1) * case.gencost_orig.c2 .* segment_width
```
rouille: Thanks!

src/model.jl
```diff
@@ -139,12 +135,12 @@ function _make_sets(
     num_solar = length(gen_solar_idx)
     num_hydro = length(gen_hydro_idx)
     # Generator cost curve segments
-    piecewise_enabled = (sum(case.gencost[:, MODEL] .== 1) > 0)
+    piecewise_enabled = (sum(case.gencost.type .== 1) > 0)
```
rouille: Thanks!

src/model.jl
```diff
@@ -139,12 +135,12 @@ function _make_sets(
...
     @assert(
         piecewise_enabled,
         "No piecewise segments detected. " * "Did you forget to linearize_gencost?"
     )
-    num_segments = convert(Int, maximum(case.gencost[:, NCOST])) - 1
+    num_segments = convert(Int, maximum(case.gencost.n)) - 1
```
rouille: Thanks!

src/model.jl
```diff
@@ -508,9 +504,7 @@ function _add_objective_function!(
     storage_e0::Array{Float64,1},
     demand_flexibility::DemandFlexibility,
 )
-    # Positional indices from mpc.gencost
-    COST = 5
-    fixed_cost = case.gencost[:, COST + 1]
+    fixed_cost = case.gencost.p1
```
rouille: Thanks!

docs/installation.rst
```diff
@@ -192,6 +192,9 @@ license file and another containing the necessary input files for the engine.
 
 .. code-block:: bash
 
-   docker run -it -v /LOCAL/PATH/TO/GUROBI.LIC:/usr/share/gurobi_license -v /LOCAL/PATH/TO/INPUT/DATA:/usr/share/data reisejl bash
+   docker run -it `
+   -v /path/to/gurobi.lic:/usr/share/gurobi_license `
+   -v /path/to/input/data:/mnt/bes/pcm `
```
BainanXia: Wonder why we change it into `/mnt/bes/pcm` here, does it have to be?

docs/usage.rst
BainanXia: I presume it is now optional and defaults to the execute directory.

docs/usage.rst
```diff
@@ -178,14 +163,15 @@ were used to run the simulation:
 
 .. code-block:: bash
 
-   pyreisejl/utility/extract_data.py -s '2016-01-01' -e '2016-01-07' -x '/PATH/TO/OUTPUT/DATA'
+   pyreisejl/utility/extract_data.py -s '2016-01-01' -e '2016-01-07' -i '/PATH/TO/INPUT/DATA'
```
BainanXia, line 162: I understand the path to output data is optional but why the input dir needs to be specified now but previously not?

pyreisejl/utility/converters.py
```diff
@@ -1,132 +1,51 @@
...
+
+    storage = grid.storage
+    if not storage["gen"].empty:
+        _save_storage(path, "StorageData", storage["StorageData"])
+        _save_storage(path, "storage_gen", storage["gen"])
```
BainanXia, line 51: @lanesmith I'm wondering how a scenario with demand flexibility enabled looks like at this stage. Currently, we have `demand_flexibility_parameters.csv` passed through with corresponding profiles in the input folder, right? Those are additional information that a scenario object carries other than `start date`, `end date` and `interval`. Should we add some checks in `call.py` regarding that as well?

pyreisejl/utility/helpers.py
```diff
@@ -167,17 +167,16 @@ def get_scenario(scenario_id):
     """Returns scenario information.
 
     :param int/str scenario_id: scenario index.
-    :return: (*tuple*) -- scenario start_date, end date, interval, input_dir, execute_dir
+    :return: (*tuple*) -- scenario start_date, end date, interval, input_dir
```
BainanXia: `end date` -> `end_date`

src/read.jl
```diff
@@ -38,8 +38,10 @@ function read_case(filepath)
     case["gen_ramp30"] = convert(Array{Float64,1}, plant.ramp_30)
 
     # Generator costs
-    gencost = DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))
-    case["gencost"] = convert(Matrix{Float64}, gencost)
+    case["gencost"] = DataFrames.DataFrame(CSV.File(joinpath(filepath, "gencost.csv")))
+    case["gencost_orig"] = DataFrames.DataFrame(
+        CSV.File(joinpath(filepath, "gencost_orig.csv"))
+    )
```
BainanXia: Leave a mark here: if we use `gencost_before` and `gencost_after` as @rouille suggested, we will need to modify these accordingly.

src/types.jl
```diff
@@ -26,8 +26,8 @@ Base.@kwdef struct Case
     gen_pmin::Array{Float64,1}
     gen_ramp30::Array{Float64,1}
 
-    gencost::Array{Float64,2}
-    gencost_orig::Array{Float64,2}
+    gencost::DataFrames.DataFrame
+    gencost_orig::DataFrames.DataFrame
```
BainanXia: Same here.

pyreisejl/utility/extract_data.py
```diff
@@ -271,60 +264,55 @@ def _cast_keys_as_lists(dictionary):
...
 
-    :param str execute_dir: directory containing all of the result.mat files from REISE.jl
+    :param str input_dir: tmp/scenario_*
     :param str start_date: the start date of the simulation run
     :param str end_date: the end date of the simulation run
     :param str scenario_id: optional identifier for the scenario, used to label output files
-    :param str output_dir: optional directory in which to store the outputs. defaults to the execute_dir
-    :param str mat_dir: optional directory in which to store the converted grid.mat file. defaults to the execute_dir
+    :param str output_dir: optional directory in which to store the outputs
+    :param str freq: the frequency of timestamps in the input profiles as a pandas frequency alias
```
BainanXia, line 305: Good catch!

BainanXia (commented):
Using column names instead of positional indices is much cleaner 💯

docs/usage.rst
rouille: If optional arguments are surrounded by brackets, then `THREADS` needs brackets too.

docs/usage.rst
```diff
@@ -135,15 +129,10 @@ The full list of arguments can be accessed via :bash:`pyreisejl/utility/call.py
                             will be deleted. The extraction process can be memory
                             intensive. This is optional and defaults to False if the
                             flag is omitted.
-      -o OUTPUT_DIR, --output-dir OUTPUT_DIR
+      -o [OUTPUT_DIR], --output-dir [OUTPUT_DIR]
                             The directory to store the extracted data. This is optional
                             and defaults to the execute directory. This flag is only
```
rouille: `execute directory` --> `input_directory`?

pyreisejl/utility/extract_data.py
```diff
@@ -333,14 +321,12 @@ def extract_scenario(
     build_log(mat_results, cost, output_dir, scenario_id)
 
     # Update outputs with date indices from the copied input.mat file
```
rouille: Remove reference to **input.mat**

docs/usage.rst
rouille: I think it should have been `-i` in place of `-x` in the previous version

docs/installation.rst
jenhagg: I had some intermediate code changes when I did this that were using the paths from const.py, but I think at this point it can be arbitrary and just has to be consistent with the command line args.

docs/usage.rst
jenhagg: ~~I think it's based on whether the `nargs` and `default` options are passed to `add_argument` when defining the parser. We don't have those for threads, but it nevertheless defaults to `None`.. I'll see if we can add that so the `--help` shows the brackets in that case.~~

edit: the brackets do come from [nargs](https://docs.python.org/3/library/argparse.html#nargs) but they represent a possible list option, like `--foo a b` gets parsed into `["a", "b"]`. The way we are using it (`nargs="?"`) is actually fine, but we don't need the extra functionality it provides so I'm gonna remove it.

pyreisejl/utility/converters.py
jenhagg: Done

docs/installation.rst
jenhagg: Done

docs/usage.rst
jenhagg: updated - default is a folder in the input directory

pyreisejl/utility/converters.py
lanesmith: Correct. All of the information pertaining to a scenario with demand flexibility enabled is contained within .csv files (e.g., `demand_flexibility_parameters.csv`, `demand_flexibility_up.csv`, etc.). Since that information is passed as .csv files, there is nothing that needs to be done to those files in this stage (i.e., `pyreisejl/utility/converters.py`). There are some checks that occur in `read.jl` for the different demand-flexibility-related files. What kind of checks did you have in mind?

BainanXia: Idk. Currently, all demand flexibility related checks are run in Julia side, not sure whether we should port it out to Python side or even `PowerSimData` (or have them in both ends), similarly as what we have done with `storage`. Or this feature so far is still engine dependent, i.e. `REISE.jl` only, and we should keep it in this way. Just want to get your thoughts, no concrete ideas yet.

lanesmith: I don't think there are any demand-flexibility-related checks that are engine dependent. Since most of the demand-flexibility-related information is profile-based and isn't really adjusted by `PowerSimData` (as of now), there hasn't really been much of a need to read those files and check that data in `PowerSimData`. I believe the other demand flexibility parameters have minor checks in `PowerSimData`, but I might be misremembering. I know we've talked about making the integration between `REISE.jl` and `PowerSimData` tighter (i.e., having `REISE.jl` be something that is only run through `PowerSimData` and not have it be run in the stand-alone version). Until we implement that, I think it makes sense to keep the checks in `REISE.jl` so that errors can be caught no matter how the engine is run.

BainanXia: By engine dependent I mean only `REISE.jl` can handle demand flexibility (eg, `REISE`, `PowerSimulationsWrapper.jl`, `PypsaWrapper`, etc), thus all related logic should be done in `REISE.jl` at this point and there is no need to factor out  engine-independent part, such as checks, into `PowerSimData`.

rouille (approved):
It looks really good. Thank you for your work, it is a considerable improvement in terms of transparency for the user (no more modification of the grid) and data management.

BainanXia (approved):
Thanks!

lanesmith (approved):
Looks good to me

PR #189, opened 2022-11-29
fix: default argument

jenhagg:

### Purpose
Fix error when running without scenario id.

### What the code is doing
Add back default specification

### Testing
Lane tested it

### Time estimate
1 min


lanesmith (approved):
Looks good, thanks!

rouille (approved):

PR #190, opened 2022-12-02
Make simulation engine more flexible to different profile-based generator types

lanesmith:

### Purpose
The purpose of this PR is to allow the simulation engine to be flexible to different types of profile-based generators. This functionality replaces the hard-coded variables and constraints that were previously associated with solar, wind, and hydro generation.

### What the code is doing
To achieve the above purpose, this code does the following:
- Accessing the plant immutable characteristics from the `grid.pkl` file, which is provided at input, and saving them to a JSON file so they can be accessed in `read.jl`.
- Setting the upper bounds of all profile-based generators to infinity so that the specified generator upper bounds (i.e., the nameplate capacities) do not interfere with the upper bounds that will be set by the provided generation profiles.
- Establishing new upper and lower bound constraints that are flexible to different types of profile-based generators. The upper bound constraint takes the value of the generation profile for each generator at each time step as the corresponding upper bound. The lower bound constraint takes the value of the generation profile for each generator at each time step multiplied by each generator type's `pmin_as_share_of_pmax` parameter (one of the plant immutable parameters) as the corresponding lower bound. These new generation constraints replace the hard-coded constraints for solar, wind, and hydro resources.
- Fixing a couple small bugs that were present in `read.jl`.

### Testing
I tested the base ERCOT case using both this updated version of `REISE.jl` and the original version of `REISE.jl` that is currently in the develop branch. They both produced the same results for a three-day simulation, as would be expected.

### Where to look
`pyreisejl/utility/call.py` and `pyreise/utility/converters.py` contain the code responsible for accessing the plant immutables from the `grid.pkl` file and saving them as a JSON file. `src/read.jl` contains the functionality to read in the plant immutables and change the specified upper bounds of the profile-based generators to infinity. `src/model.jl` and `src/loop.jl` contain the functionality for setting and updating the described constraints.

### Time estimate
The changes to all files other than `model.jl` and `loop.jl` will hopefully seem pretty straightforward and quick to review. `model.jl` and `loop.jl` feature the bulk of the new functionality, but hopefully shouldn't take too long.


pyreisejl/utility/converters.py
```diff
@@ -49,3 +50,26 @@ def pkl_to_csv(path):
     if not storage["gen"].empty:
         _save_storage(path, "StorageData", storage["StorageData"])
         _save_storage(path, "storage_gen", storage["gen"])
+
+
+def pkl_to_json(path):
```
BainanXia: This is nice. Just wonder why we need `json` instead of `csv` here.

pyreisejl/utility/converters.py
```diff
@@ -49,3 +50,26 @@ def pkl_to_csv(path):
...
+    # Convert sets to lists so a .json file can be created
+    class SetEncoder(json.JSONEncoder):
+        def default(self, obj):
+            if isinstance(obj, set):
+                return list(obj)
+            return json.JSONEncoder.default(self, obj)
+
+    # Access the grid object from the .pkl file
+    with open(os.path.join(path, "grid.pkl"), "rb") as f:
+        grid = pickle.load(f)
```
BainanXia: Should we extract the loading part out so that we only do it once during `grid.pkl` to `csv` or `json`?

src/model.jl
```diff
@@ -155,29 +156,42 @@ function _make_sets(
...
     num_segments = convert(Int, maximum(case.gencost_after.n)) - 1
     segment_idx = 1:num_segments
+
+    # Create index for different profile-based generators
+    profile_resources_idx = Dict{String,Array{Int64,1}}()
+    for g in case.profile_resources
+        profile_resources_idx[g] = gen_idx[findall(case.genfuel .== g)]
+    end
+
+    # Create numerical representation of the different profile-based generators
```
BainanXia: Not sure whether it is more clear to say `#Create index for different profile-based resource types` here

src/loop.jl
```diff
@@ -109,28 +100,54 @@ function interval_loop(
...
+                        )
+                        JuMP.set_normalized_rhs(
+                            m[:profile_lower_bound][g, i, h],
+                            case.pmin_as_share_of_pmax[sets.profile_resources_num_rep[g]] *
+                            simulation_profile[sets.profile_to_group[sets.profile_resources_num_rep[g]]][
+                                h, i
+                            ],
+                        )
+                    end
+                end
```
BainanXia, line 131: Much cleaner!

src/read.jl
```diff
@@ -50,6 +56,15 @@ function read_case(filepath)
...
     )
 
+    # Set the PMAX for all profile-based generators to Inf; the true PMAX for profile-
+    # based generators will be determined by the provided profile
+    for g in unique(case["genfuel"])
+        gen_idx = case["genfuel"] .== g
+        if g in case["profile_resources"]
+            case["gen_pmax"][gen_idx] .= Inf
+        end
+    end
```
BainanXia: Not sure whether it is more straightforward to do
```julia
for g in case["profile_resources"]
    case["gen_pmax"][case["genfuel"] .== g] .= Inf
end
```

BainanXia (approved):
Nice and neat. Thanks.

pyreisejl/utility/converters.py
lanesmith: The plant immutables are saved in the `grid.pkl` file as a dictionary, so JSON seemed to be the way to go.

pyreisejl/utility/converters.py
lanesmith: That's a good idea. I'll implement that.

src/model.jl
lanesmith: That is more clear. I'll change it.

src/loop.jl
lanesmith: Thanks!

src/read.jl
lanesmith: That should work. I'll make sure and implement it.

pyreisejl/utility/converters.py
lanesmith: Done.

src/read.jl
lanesmith: Done.

src/model.jl
lanesmith: Done.

BainanXia (approved):
LGTM. Thanks!

rouille (approved):
🙏

PR #194, opened 2023-02-02
fix: use dataframe for storage.gen

jenhagg:

### Purpose
Loosen restriction on column types for storage (specifically string bus_id) to handle both usa_tamu and europe_tub grids. 

### What the code is doing
Change `storage.gen` to a data frame, rather than attempting conversion to a float array. Also filter the columns we extract from the original `grid.pkl` to just the ones we use (bus_id, Pmin, Pmax). 

### Testing
Ran a usa_tamu scenario with storage, ensured storage is loaded and results files include storage.

### Time estimate
5 min

### DIFF
--- a/pyreisejl/utility/converters.py
+++ b/pyreisejl/utility/converters.py
@@ -3,26 +3,14 @@ import os
 import pickle
 
 cols = {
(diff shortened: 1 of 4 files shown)


rouille (approved):
Much better. Thanks

BainanXia (approved):
Very nice. Thanks!

PR #195, opened 2023-02-09
refactor: reduce indexing by using values directly

jenhagg:
### Purpose
Readability improvement. 

### What the code is doing
Simplify the constraints by using julia's builtin `values()` function to remove the extra dict lookup. 

### Testing
Ran a 4 day europe pcm which completed successfully.

### Time estimate
5 min

### DIFF
--- a/src/loop.jl
+++ b/src/loop.jl
@@ -106,24 +106,17 @@ function interval_loop(
                     getfield(case, Symbol(p))[interval_start:interval_end, 2:end]
                 )
             end
-            for g in keys(sets.profile_resources_num_rep)
+            for g in values(sets.profile_resources_num_rep)
                 for h in 1:interval
-                    for i in
-                        1:length(
-                        sets.profile_resources_idx[sets.profile_resources_num_rep[g]]
-                    )
+                    for i in 1:length(sets.profile_resources_idx[g])
                         JuMP.set_normalized_rhs(
                             m[:profile_upper_bound][g, i, h],
-                            simulation_profile[sets.profile_to_group[sets.profile_resources_num_rep[g]]][
-                                h, i
-                            ],
+                            simulation_profile[sets.profile_to_group[g]][h, i],
                         )
                         JuMP.set_normalized_rhs(
                             m[:profile_lower_bound][g, i, h],
-                            case.pmin_as_share_of_pmax[sets.profile_resources_num_rep[g]] *
-                            simulation_profile[sets.profile_to_group[sets.profile_resources_num_rep[g]]][
-                                h, i
-                            ],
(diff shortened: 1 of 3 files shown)


src/loop.jl
```diff
@@ -106,24 +106,17 @@ function interval_loop(
...
-                        1:length(
-                        sets.profile_resources_idx[sets.profile_resources_num_rep[g]]
-                    )
+                    for i in 1:length(sets.profile_resources_idx[g])
                         JuMP.set_normalized_rhs(
                             m[:profile_upper_bound][g, i, h],
-                            simulation_profile[sets.profile_to_group[sets.profile_resources_num_rep[g]]][
-                                h, i
-                            ],
+                            simulation_profile[sets.profile_to_group[g]][h, i],
```
jenhagg, line 114: Not sure for the above line `m[:profile_upper_bound][g, i, h]` whether the `g` index is arbitrary, but since that is being changed, could someone confirm that is the case?

src/model.jl
```diff
@@ -542,28 +542,26 @@ function _add_profile_generator_limits!(
     JuMP.@constraint(
         m,
         profile_upper_bound[
-            g in keys(sets.profile_resources_num_rep),
-            i in 1:length(sets.profile_resources_idx[sets.profile_resources_num_rep[g]]),
+            g in values(sets.profile_resources_num_rep),
```
lanesmith: `JuMP` is okay with using strings as indices instead of integers? If so, thanks for catching that! My original formulation is a lot grosser because I thought you needed to use integers to index the constraints.

src/loop.jl
lanesmith: So long as `JuMP` constraints are okay with using strings as indices, I think this looks right. The `g` index is not arbitrary, it is helping to specify the specific constraint that is being updated. We are essentially creating a constraint for each profile-based generator type (this is what `g` is referring to), each generator of each profile-based generator type, and each hourly time step.

src/loop.jl
```diff
@@ -106,24 +106,17 @@ function interval_loop(
                     getfield(case, Symbol(p))[interval_start:interval_end, 2:end]
                 )
             end
-            for g in keys(sets.profile_resources_num_rep)
+            for g in values(sets.profile_resources_num_rep)
```
lanesmith: I think we can just use `for g in case.profile_resources` instead. If so, I believe `profile_resources_num_rep` can be removed from `sets`.

src/loop.jl
rouille: I guess it is the case if @jenhagg was able to run a PCM for Europe, no?

lanesmith (approved):
Looks good to me. Thanks!

PR #196, opened 2023-02-22
Generalize type handling to support europe grid

jenhagg:

### Purpose
Make it so we can handle string or int for certain columns in the grid. This enables using both europe_tub and usa_tamu grid models. 

### What the code is doing
Use a union type when reading in grid csv files, and change the index of `bus_id2idx` to always be string. I'm sure there is a more elegant way to do this, but wanted to get something merged that works.

### Testing
Ran a simulation using usa_tamu, and the 128 node europe grid.

### Time estimate
5 min


BainanXia (approved):
It makes sense.

lanesmith (approved):
Looks good to me. Thanks!

